PR #32, opened 2021-11-04
create a snapshot before creating tar archive

schwannden:
if during backup, the content of the backup folder continues to change, tar returns a fatal error
```
level=error msg="Fatal error running backup: takeBackup: error compressing backup folder: archive/
tar: write too long"
```

this is because tar create a header containing files to be compressed, and write their size in header. But in actual compressing, it find file size has changed.

If we stop the container during backup, this situation should not happen, however in my scenario, it is desirable that we keep container running. Unfortunately this results in occasional backup failure due to the continuously growing logs files in my backup folder.

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,12 +126,12 @@ You can populate below template according to your requirements and use it as you
 
 # BACKUP_LATEST_SYMLINK="backup.latest.tar.gz"
 
-# Wheter to copy the content of backup folder before performing archive.
+# Wehter to copy the content of backup folder before creating archive.
 # In rare scenario where the content of the source backup volume is still
 # updating, but we do not wish to stop the container while performing backup,
-# this snapshot before archive ensures the integrey of the tar.gz file.
+# this snapshot before archive ensures the integrity of the tar.gz file.
 
-# BACKUP_SNAPSHOT_BEFORE_ARCHIVE="false"
+# BACKUP_FROM_SNAPSHOT="false"
 
 ########### BACKUP STORAGE
 
--- a/README.md
+++ b/README.md
@@ -126,6 +126,13 @@ You can populate below template according to your requirements and use it as you
 
 # BACKUP_LATEST_SYMLINK="backup.latest.tar.gz"
 
+# Wheter to copy the content of backup folder before performing archive.
+# In rare scenario where the content of the source backup volume is still
+# updating, but we do not wish to stop the container while performing backup,
+# this snapshot before archive ensures the integrey of the tar.gz file.
+
+# BACKUP_SNAPSHOT_BEFORE_ARCHIVE="false"
+
 ########### BACKUP STORAGE
 
 # The name of the remote bucket that should be used for storing backups. If
--- a/cmd/backup/main.go
+++ b/cmd/backup/main.go
@@ -328,32 +328,27 @@ func (s *script) stopContainers() (func() error, error) {
 	}, stopError
 }
 
+func (s *script) snapshotFolder() string {
+	return filepath.Join("/tmp", s.c.BackupSources)
+}
+
 // takeBackup creates a tar archive of the configured backup location and
 // saves it to disk.
 func (s *script) takeBackup() error {
 	s.file = timeutil.Strftime(&s.start, s.file)
 	backupSources := s.c.BackupSources
 	if s.c.BackupFromSnapshot {
-		backupCopiedSource := filepath.Join("/tmp", s.c.BackupSources)
-		backupSources = backupCopiedSource
-		s.logger.Infof("Created snapshot of `%s` at `%s`.", s.c.BackupSources, backupCopiedSource)
+		backupSources = s.snapshotFolder()
+		s.logger.Infof("Created snapshot of `%s` at `%s`.", s.c.BackupSources, backupSources)
 		// copy before compressing guard against a situation where backup folder's content are still growing.
-		if err := copy.Copy(s.c.BackupSources, backupCopiedSource, copy.Options{ PreserveTimes: true }); err != nil {
+		if err := copy.Copy(s.c.BackupSources, backupSources, copy.Options{ PreserveTimes: true }); err != nil {
 			return fmt.Errorf("takeBackup: error creating snapshot on backup folder: %w", err)
 		}
-		defer func() {
-			if err := os.RemoveAll(backupCopiedSource); err != nil {
-				s.logger.Errorf("takeBackup: error removing snapshot folder %s.", backupCopiedSource)
-				os.Exit(1)
-			}
-			s.logger.Infof("Removed snapshot folder %s.", backupCopiedSource)
-		}()
 	}
 	if err := targz.Compress(backupSources, s.file); err != nil {
 		return fmt.Errorf("takeBackup: error compressing backup folder: %w", err)
 	}
 	s.logger.Infof("Created backup of `%s` at `%s`.", backupSources, s.file)
-	s.logger.Infof("Removed local artifacts %s.", s.file)
 	return nil
 }
 
@@ -430,8 +425,15 @@ func (s *script) copyBackup() error {
 	return nil
 }
 
-// removeArtifacts removes the backup file from disk.
+// removeArtifacts removes the backup file from disk. Also remove snapshot if snapshot was created
 func (s *script) removeArtifacts() error {
+	if s.c.BackupFromSnapshot {
+		snapshotFolder := s.snapshotFolder()
+		if err := os.RemoveAll(snapshotFolder); err != nil {
+			return fmt.Errorf("removeArtifacts: error removing snapshot folder %s: %w", snapshotFolder, err)
+		}
+		s.logger.Infof("Removed snapshot folder %s.", snapshotFolder)
+	}
 	_, err := os.Stat(s.file)
 	if err != nil {
 		if os.IsNotExist(err) {
--- a/cmd/backup/main.go
+++ b/cmd/backup/main.go
@@ -96,6 +96,7 @@ type config struct {
 	BackupPruningLeeway        time.Duration `split_words:"true" default:"1m"`
 	BackupPruningPrefix        string        `split_words:"true"`
 	BackupStopContainerLabel   string        `split_words:"true" default:"true"`
+	BackupFromSnapshot         bool          `split_words:"true"`
 	AwsS3BucketName            string        `split_words:"true"`
 	AwsEndpoint                string        `split_words:"true" default:"s3.amazonaws.com"`
 	AwsEndpointProto           string        `split_words:"true" default:"https"`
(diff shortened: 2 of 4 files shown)


go.mod
```diff
@@ -4,11 +4,13 @@ go 1.17
 
 require (
 	github.com/docker/docker v20.10.8+incompatible
+	github.com/go-gomail/gomail v0.0.0-20160411212932-81ebce5c23df
```
schwannden, line 7: no entirely sure if updating this is ok? just realized this module is updated along the way. If you don't want this update I can remove this.

m90: It's still the same version so I guess it's fine. Not sure why it ever has been `indirect` tbh as it's used by the script itself.

cmd/backup/main.go
```diff
@@ -330,10 +331,21 @@ func (s *script) stopContainers() (func() error, error) {
...
+	if err := copy.Copy(s.c.BackupSources, backupCopiedSource); err != nil {
+		return fmt.Errorf("takeBackup: error creating snapshot on backup folder: %w", err)
+	}
+	if err := targz.Compress(backupCopiedSource, s.file); err != nil {
 		return fmt.Errorf("takeBackup: error compressing backup folder: %w", err)
 	}
 	s.logger.Infof("Created backup of `%s` at `%s`.", s.c.BackupSources, s.file)
+	s.logger.Infof("Removed local artifacts %s.", s.file)
+
+	if err := os.RemoveAll(backupCopiedSource); err != nil {
```
m90, line 345: Can we `defer` this as we also want to run it in case compressing fails or similar?

cmd/backup/main.go
```diff
@@ -330,10 +331,21 @@ func (s *script) stopContainers() (func() error, error) {
 // saves it to disk.
 func (s *script) takeBackup() error {
 	s.file = timeutil.Strftime(&s.start, s.file)
-	if err := targz.Compress(s.c.BackupSources, s.file); err != nil {
+  backupCopiedSource := filepath.Join("/tmp", s.c.BackupSources)
+	// copy before compressing guard against a situation where backup folder's content are still growing.
+	if err := copy.Copy(s.c.BackupSources, backupCopiedSource); err != nil {
```
m90, line 336: Two questions about the behavior of this library:

- what happens if there are symlinks in the copied file tree? Do they get resolved?
- what happens if files in `s.c.BackupSources` change while copying is done? (i.e. how does this not fail in such a scenario)

m90 (commented):
Thanks for thinking of this.

I'm wondering if we should maybe make this behavior opt-in because of the following reason: I think the majority of users (including myself) will stop their containers when taking a backup, so creating the archive from a snapshot is not really necessary. However, if there are a lot of files to be copied, this can become a lengthy operation which - in the case of stopped containers - makes the downtime caused by a backup run even longer for no good reason. Stopping your containers during backup should still be the advised procedure.

cmd/backup/main.go
schwannden: 1. no symlinks are not copied by default (https://github.com/otiai10/copy/blob/edeb9f37d4d0c28725014fcab6635120197932a1/copy.go#L163).
2. since we are compressing the snapshot folder (copied folder) now, so even if the content of s.c.BackupSources is grwoing, it does not affect the snapshot folder (copied folder). So the archived content is the content in the moment copy finish. anything added after that is not archived.

cmd/backup/main.go
schwannden: yeah this is a good point I will fix this

schwannden:
> Thanks for thinking of this.
> 
> I'm wondering if we should maybe make this behavior opt-in because of the following reason: I think the majority of users (including myself) will stop their containers when taking a backup, so creating the archive from a snapshot is not really necessary. However, if there are a lot of files to be copied, this can become a lengthy operation which - in the case of stopped containers - makes the downtime caused by a backup run even longer for no good reason. Stopping your containers during backup should still be the advised procedure.

sure can do!

cmd/backup/main.go
m90: > no symlinks are not copied by default

From reading the code it seems to me `Shallow` is used by default, but that's good as it would be the exact same behavior as before: https://github.com/otiai10/copy/blob/edeb9f37d4d0c28725014fcab6635120197932a1/options.go#L70-L72

m90:
Maybe `BACKUP_FROM_SNAPSHOT` would be a good option name, what do you think?

cmd/backup/main.go
m90: I also wonder if we should use the `PreserveTimes` option here as otherwise all of the copied files would lose their timestamps (which might be important for some consumers): https://github.com/otiai10/copy/tree/edeb9f37d4d0c28725014fcab6635120197932a1#advanced-usage

schwannden: I see, will add that!

cmd/backup/main.go
```diff
@@ -330,10 +332,28 @@ func (s *script) stopContainers() (func() error, error) {
...
+	backupSources := s.c.BackupSources
+	if s.c.BackupFromSnapshot {
+		backupCopiedSource := filepath.Join("/tmp", s.c.BackupSources)
+		backupSources = backupCopiedSource
+		s.logger.Infof("Created snapshot of `%s` at `%s`.", s.c.BackupSources, backupCopiedSource)
+		// copy before compressing guard against a situation where backup folder's content are still growing.
+		if err := copy.Copy(s.c.BackupSources, backupCopiedSource, copy.Options{ PreserveTimes: true }); err != nil {
+			return fmt.Errorf("takeBackup: error creating snapshot on backup folder: %w", err)
+		}
+		defer func() {
```
schwannden, line 344: so honestly I have never written go before and idk if this is what you are thinking, I tested it and it seems ok. Guess I really need to learn more about go if I am going to continue to make pr here

m90: That's exactly how defer works :) no need to worry.

cmd/backup/main.go
```diff
@@ -330,10 +332,28 @@ func (s *script) stopContainers() (func() error, error) {
...
+			if err := os.RemoveAll(backupCopiedSource); err != nil {
+				s.logger.Errorf("takeBackup: error removing snapshot folder %s.", backupCopiedSource)
+				os.Exit(1)
```
m90, line 347: This is correct because we don't want to continue running the backup script anymore, but it also creates a different problem: everything else that is still "to unwind" (Go jargon for running things that are deferred before exiting) will never run in this case as the program immediately exits. This happens to include the hooks for sending out notifications on failed backups which is a little unfortunate (as this is a failed backup).

I think we have two options for fixing this:

### Make this a warning only

We could tolerate this error, log it as a warning and let the script continue to run. This could create a few weird edgecases but would probably be ok in most cases.

### Have `func main` handle all orchestration

This program already has a pattern for handling such cases, which is having a method return a cleanup function that is called no matter what. `func main` then schedules this function to run no matter if an error was returned or not.

If we wanted to apply this pattern here as well, we'd need to do the following:
- Change the signature of `(*script).takeBackup` to be `takeBackup() (func() error, error)`
- Instead of deferring the function here, return it (or return a noop in case the feature is not used)
- When calling the method in `func main` it would look something like this:
    ```go
	cleanup, err := s.takeBackup()
	defer func() {
		s.must(cleanup())
	}()
	if err != nil {
		return err
	}
   ```

The program already follows this pattern when stopping containers (which returns a function to bring them back up), so you can also check how this looks like there.

---

I would probably prefer using option 2 as it's the most correct, but it's also making this more complicated. If you feel like this is too much, let me know. Also sorry for not thinking about this in depth with my initial comment.

README.md
```diff
@@ -126,6 +126,13 @@ You can populate below template according to your requirements and use it as you
 
 # BACKUP_LATEST_SYMLINK="backup.latest.tar.gz"
 
+# Wheter to copy the content of backup folder before performing archive.
+# In rare scenario where the content of the source backup volume is still
+# updating, but we do not wish to stop the container while performing backup,
+# this snapshot before archive ensures the integrey of the tar.gz file.
```
m90, line 132:
```suggestion
# this snapshot before archive ensures the integrity of the tar.gz file.
```

README.md
```diff
@@ -126,6 +126,13 @@ You can populate below template according to your requirements and use it as you
...
+
+# BACKUP_SNAPSHOT_BEFORE_ARCHIVE="false"
```
m90, line 134:
```suggestion
# BACKUP_FROM_SNAPSHOT="false"
```

m90 (commented):
It would be great if we could also include this feature in one of the tests.

The `test/cli/run.sh` test does currently stop containers, but it's also never checking whether they are actually stopped, so how about we remove the label here and start using the `BACKUP_FROM_SNAPSHOT` option so we have this portion of the code covered as well. All assertions the test is making should still pass as-is. In case you prefer to skip this I can also take care of this post-merging.

Thanks :+1:

cmd/backup/main.go
```diff
@@ -330,10 +332,28 @@ func (s *script) stopContainers() (func() error, error) {
...
+			}
+			s.logger.Infof("Removed snapshot folder %s.", backupCopiedSource)
+		}()
+	}
+	if err := targz.Compress(backupSources, s.file); err != nil {
 		return fmt.Errorf("takeBackup: error compressing backup folder: %w", err)
 	}
-	s.logger.Infof("Created backup of `%s` at `%s`.", s.c.BackupSources, s.file)
+	s.logger.Infof("Created backup of `%s` at `%s`.", backupSources, s.file)
+	s.logger.Infof("Removed local artifacts %s.", s.file)
```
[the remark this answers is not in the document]
schwannden: this got accidentally pasted here :p

cmd/backup/main.go
schwannden: is it possible if I simply move my cleanup function to `removeArtifacts` function?
```
func (s *script) removeArtifacts() error {
	_, err := os.Stat(s.file)
	if err != nil {
		if os.IsNotExist(err) {
			return nil
		}
		return fmt.Errorf("removeArtifacts: error calling stat on file %s: %w", s.file, err)
	}
	if err := os.Remove(s.file); err != nil {
		return fmt.Errorf("removeArtifacts: error removing file %s: %w", s.file, err)
	}
	s.logger.Infof("Removed local artifacts %s.", s.file)
  if s.c.BackupFromSnapshot {
    if err := os.RemoveAll(backupSources); err != nil {
      return fmt.Errorf("takeBackup: error removing snapshot folder %s.", backupSources)
    }
    s.logger.Infof("Removed snapshot folder %s.", backupSources)
  }

	return nil
}
```

cmd/backup/main.go
```diff
@@ -430,8 +425,15 @@ func (s *script) copyBackup() error {
 	return nil
 }
 
-// removeArtifacts removes the backup file from disk.
+// removeArtifacts removes the backup file from disk. Also remove snapshot if snapshot was created
 func (s *script) removeArtifacts() error {
+	if s.c.BackupFromSnapshot {
```
schwannden, line 430: I put remove snapshot before remove artifact, reverse to the archive creation order in `takeBackup`. This has the benefit that if creating snapshot already failed before creating archive, snapshot still get removed here.

schwannden:
> It would be great if we could also include this feature in one of the tests.
> 
> The `test/cli/run.sh` test does currently stop containers, but it's also never checking whether they are actually stopped, so how about we remove the label here and start using the `BACKUP_FROM_SNAPSHOT` option so we have this portion of the code covered as well. All assertions the test is making should still pass as-is. In case you prefer to skip this I can also take care of this post-merging.
> 
> Thanks +1

I don't quite understand this part? What do you mean by `remove the label here`, which label are you referring to?

cmd/backup/main.go
schwannden: thx for the lengthy explanation, this is helpful for me to understand this project better

m90: > is it possible if I simply move my cleanup function to removeArtifacts function?

That's a good middle ground, I like it. In the long run such clean up could be moved into an "all-levels" exit hook, but I'd have to think about this further. For now, this solution is just right :+1:

m90:
> I don't quite understand this part? What do you mean by remove the label here, which label are you referring to?

Sorry for not being clear enough about the tests. The "cli" test creates a container/volume that should be backed up and applies a `docker-volume-backup.stop-during-backup=true` label to the container here: https://github.com/offen/docker-volume-backup/blob/0c666d0c880a8f621c72010fdf8446a0e6cda11c/test/cli/run.sh#L23-L28

This means, that when the backup is run further below: https://github.com/offen/docker-volume-backup/blob/0c666d0c880a8f621c72010fdf8446a0e6cda11c/test/cli/run.sh#L32-L43

the container will be stopped and restarted. As we also want to test the opposite behavior now, all that would be needed to do is:

- create the long-running container without the label
- when running the backup, pass `BACKUP_FROM_SNAPSHOT=true`

m90:
@schwannden The changes look good now, but I'd still like to include the newly introduced option in a test (see above). Do you think you could still get this set up? In case you're busy and would prefer not to do it, it's also fine and I can do it post-merge. Just let me know what you prefer and we can get this feature released. Thanks.

schwannden:
> @schwannden The changes look good now, but I'd still like to include the newly introduced option in a test (see above). Do you think you could still get this set up? In case you're busy and would prefer not to do it, it's also fine and I can do it post-merge. Just let me know what you prefer and we can get this feature released. Thanks.
@m90  thx for reminding and sorry work has been a bit caught up. I understand what you said about testing now, but I guess we can simply merge it for now. I am not sure when in this week would I be able to work on the test.

m90:
Ok, great, I will take care of this myself today then and cut a release when it's working. Thanks again for your contribution!

m90:
This feature is now included in [v2.4.0](https://github.com/offen/docker-volume-backup/releases/tag/v2.4.0) - thanks again :tada:

cmd/backup/main.go
m90: I did some refactoring on this topic in here https://github.com/offen/docker-volume-backup/pull/33 in case you're interested.

PR #56, opened 2022-02-04
Updated README.md "Restoring..." section example

Fridgemagnet:
Edited the "Restoring a volume from a backup" section example to be able to better differentiate between the names of the temporary restore container and the name of the mounted restore volume.
New example container name: temp_restore_container
The restore volume name remains the same: backup_restore

Also changed "one-off container" to "once-off container".


m90 (approved):
Congrats and thanks a lot :100:

Issue #59, opened 2022-02-08
Custom notification messages


MMauro94:
* **I'm submitting a ...**
  - [ ] bug report
  - [x] feature request
  - [ ] support request

Right now there is no way of customizing error and success messages for notifications. As I have multiple backup containers, I struggle to identify at a glance which backup has run, since I have to read the log. Also I don't need the full log on success notifications.

It would be nice having the ability to customize these messages, while injecting useful information. Something like this:

```
SUCCESS_NOTIFICATION_MESSAGE={CONTAINER_LABEL} backup succeeded at {BACKUP_TIME}
```

Thoughts? If you're interested in implementing this feature I can produce a list of useful variables to be injected.

Open questions:
1. Should the same feature be available for the notification title?
2. Should it be possible to provide different messages for different shoutrrr URLs? (For example send the full log via email but only a short message in Telegarm) For this one I'd argue no, since it would become too complex to configure by environment variables, but I just thought to put it out here.

m90:
Thanks for submitting this idea. While I probably would not use it myself I can see how it can be useful, so I would be happy if anyone wants to implement it and am happy to help with reviewing/merging PRs.

---

A few thoughts on the implementation from my end: Golang has a `text/template` package in the standard libary which would be a good fit for this I guess. Right now a notification is just hardcoded here: https://github.com/offen/docker-volume-backup/blob/3bb99a711754ba53f549c70c8d0914b6007e329d/cmd/backup/main.go#L286-L296

This could however be turned into a standalone template file something like this (I have not used the package in ages, so this might be slightly wrong syntactically, but you will get the idea):

```
{{ define "title_failure" }}
Failure running docker-volume-backup at {{ .StartTime }}
{{ end }}

{{ define "body_failure" }}
Running docker-volume-backup failed with error: {{ .Error }}
Log output of the failed run was:

{{ .LogOutput }}
{{ end }}

// this would also define `title_success` and `body_success`
```

This file could be read at runtime from e.g. `/etc/volume-backup.d/00notifications.tmpl`. Users who wanted to customize these messages could now mount a `/etc/volume-backup.d/10override.tmpl` (either at runtime or by inheriting from the Docker image) file that can selectively overwrite any of the templates defined in the default templates.

The init procedure of the command would then read and parse all files in `/etc/volume-backup.d` and use the last occurrence of each template definition.

Do you think this makes sense?


About your questions:
1. yes, definitely
2. i agree this would become to complicated to configure

[...]

PR #60, opened 2022-02-10
Added custom notification messages using text/template

MMauro94:
This should take care of #59. This is my first time ever with Golang so I'm sorry if I've done anything not in the correct way.

As per your suggestion, I've used the `text/template` package. The default template uses the exact same wording of current notifications.

A few things:
1. I've trimmed newlines on the template result, otherwise the templates has to look ugly
2. Currently supported params for templates: `StartTime`, `LogOutput`. `BackupStopContainerLabel` (and `Error` for failures). I would have liked to add more (like number of stopped containers, number of pruned backups, took time, and many more), but some things would have required too much refactor of the code to have this variables ready in the `notifySuccess` function. Any thoughts on this?
3. I've yet to update the README, waiting on your feedback.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,8 +6,7 @@ FROM golang:1.17-alpine as builder
 WORKDIR /app
 COPY go.mod go.sum ./
 RUN go mod download
-COPY cmd/backup/notifications.tmpl ./cmd/backup/notifications.tmpl
-COPY cmd/backup/main.go ./cmd/backup/main.go
+COPY cmd/backup ./cmd/backup/
 RUN go build -o backup cmd/backup/main.go
 
 FROM alpine:3.15
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,6 +6,7 @@ FROM golang:1.17-alpine as builder
 WORKDIR /app
 COPY go.mod go.sum ./
 RUN go mod download
+COPY cmd/backup/notifications.tmpl ./cmd/backup/notifications.tmpl
 COPY cmd/backup/main.go ./cmd/backup/main.go
 RUN go build -o backup cmd/backup/main.go
 
@@ -18,7 +19,6 @@ RUN apk add --update ca-certificates
 COPY --from=builder /app/backup /usr/bin/backup
 
 COPY ./entrypoint.sh /root/
-COPY notifications.tmpl /etc/dockervolumebackup/notifications.d/00defaults.tmpl
 RUN chmod +x entrypoint.sh
 
 ENTRYPOINT ["/root/entrypoint.sh"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,7 +18,7 @@ RUN apk add --update ca-certificates
 COPY --from=builder /app/backup /usr/bin/backup
 
 COPY ./entrypoint.sh /root/
-COPY ./00notifications.tmpl /etc/volume-backup.d/
+COPY notifications.tmpl /etc/dockervolumebackup/notifications.d/00defaults.tmpl
 RUN chmod +x entrypoint.sh
 
 ENTRYPOINT ["/root/entrypoint.sh"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,6 +18,7 @@ RUN apk add --update ca-certificates
 COPY --from=builder /app/backup /usr/bin/backup
 
 COPY ./entrypoint.sh /root/
+COPY ./00notifications.tmpl /etc/volume-backup.d/
 RUN chmod +x entrypoint.sh
 
 ENTRYPOINT ["/root/entrypoint.sh"]
--- a/cmd/backup/main.go
+++ b/cmd/backup/main.go
@@ -112,26 +112,26 @@ type script struct {
 
 // ContainersStats stats about the docker containers
 type ContainersStats struct {
-	All        int
-	ToStop     int
-	Stopped    int
-	StopErrors int
+	All        uint
+	ToStop     uint
+	Stopped    uint
+	StopErrors uint
 }
 
 // BackupFileStats stats about the created backup file
 type BackupFileStats struct {
 	Name      string
 	FullPath  string
-	SizeBytes int64
+	SizeBytes uint64
 	SizeBin   string
 	SizeDec   string
 }
 
 // ArchiveStats stats about the status of an archival directory
 type ArchiveStats struct {
-	Total       int
-	Pruned      int
-	PruneErrors int
+	Total       uint
+	Pruned      uint
+	PruneErrors uint
 }
 
 // ArchivesStats stats about each possible archival location (Local, WebDAV, S3)
@@ -472,9 +472,9 @@ func (s *script) stopContainers() (func() error, error) {
 	}
 
 	s.stats.Containers = ContainersStats{
-		All:     len(allContainers),
-		ToStop:  len(containersToStop),
-		Stopped: len(stoppedContainers),
+		All:     uint(len(allContainers)),
+		ToStop:  uint(len(containersToStop)),
+		Stopped: uint(len(stoppedContainers)),
 	}
 
 	return func() error {
@@ -625,7 +625,7 @@ func (s *script) copyBackup() error {
 	} else {
 		size := stat.Size()
 		s.stats.BackupFile = BackupFileStats{
-			SizeBytes: size,
+			SizeBytes: uint64(size),
 			SizeBin:   bytesToString(size, false),
 			SizeDec:   bytesToString(size, true),
 			Name:      name,
@@ -713,8 +713,8 @@ func (s *script) pruneOldBackups() error {
 		}
 
 		s.stats.Archives.S3 = ArchiveStats{
-			Total:  lenCandidates,
-			Pruned: len(matches),
+			Total:  uint(lenCandidates),
+			Pruned: uint(len(matches)),
 		}
 		if len(matches) != 0 && len(matches) != lenCandidates {
 			objectsCh := make(chan minio.ObjectInfo)
@@ -731,7 +731,7 @@ func (s *script) pruneOldBackups() error {
 					removeErrors = append(removeErrors, result.Err)
 				}
 			}
-			s.stats.Archives.S3.PruneErrors = len(removeErrors)
+			s.stats.Archives.S3.PruneErrors = uint(len(removeErrors))
 
 			if len(removeErrors) != 0 {
 				return fmt.Errorf(
@@ -774,8 +774,8 @@ func (s *script) pruneOldBackups() error {
 		}
 
 		s.stats.Archives.WebDAV = ArchiveStats{
-			Total:  lenCandidates,
-			Pruned: len(matches),
+			Total:  uint(lenCandidates),
+			Pruned: uint(len(matches)),
 		}
 		if len(matches) != 0 && len(matches) != lenCandidates {
 			var removeErrors []error
@@ -786,7 +786,7 @@ func (s *script) pruneOldBackups() error {
 					s.logger.Infof("Pruned %s from WebDAV: %s", match.Name(), filepath.Join(s.c.WebdavUrl, s.c.WebdavPath))
 				}
 			}
-			s.stats.Archives.WebDAV.PruneErrors = len(removeErrors)
+			s.stats.Archives.WebDAV.PruneErrors = uint(len(removeErrors))
 			if len(removeErrors) != 0 {
 				return fmt.Errorf(
 					"pruneOldBackups: %d error(s) removing files from remote storage: %w",
@@ -855,8 +855,8 @@ func (s *script) pruneOldBackups() error {
 		}
 
 		s.stats.Archives.Local = ArchiveStats{
-			Total:  len(candidates),
-			Pruned: len(matches),
+			Total:  uint(len(candidates)),
+			Pruned: uint(len(matches)),
 		}
 		if len(matches) != 0 && len(matches) != len(candidates) {
 			var removeErrors []error
@@ -866,7 +866,7 @@ func (s *script) pruneOldBackups() error {
 				}
 			}
 			if len(removeErrors) != 0 {
-				s.stats.Archives.Local.PruneErrors = len(removeErrors)
+				s.stats.Archives.Local.PruneErrors = uint(len(removeErrors))
 				return fmt.Errorf(
 					"pruneOldBackups: %d error(s) deleting local files, starting with: %w",
 					len(removeErrors),
--- a/cmd/backup/main.go
+++ b/cmd/backup/main.go
@@ -83,6 +83,8 @@ func main() {
 
 	s.must(s.encryptBackup())
 	s.must(s.copyBackup())
+	s.stats.EndTime = time.Now()
+	s.stats.TookTime = s.stats.EndTime.Sub(s.stats.EndTime)
 	s.must(s.pruneOldBackups())
 }
 
@@ -98,14 +100,62 @@ type script struct {
 	hooks        []hook
 	hookLevel    hookLevel
 
-	start  time.Time
-	file   string
-	output *bytes.Buffer
+	file  string
+	stats *Stats
 
-	c *config
+	c *Config
 }
 
-type config struct {
+// ContainersStats stats about the docker containers
+type ContainersStats struct {
+	All        int
+	ToStop     int
+	Stopped    int
+	StopErrors int
+}
+
+// BackupFileStats stats about the created backup file
+type BackupFileStats struct {
+	Name      string
+	FullPath  string
+	SizeBytes int64
+	SizeBin   string
+	SizeDec   string
+}
+
+// ArchiveStats stats about the status of an archival directory
+type ArchiveStats struct {
+	Total       int
+	Pruned      int
+	PruneErrors int
+}
+
+// ArchivesStats stats about each possible archival location (Local, WebDAV, S3)
+type ArchivesStats struct {
+	Local  ArchiveStats
+	WebDAV ArchiveStats
+	S3     ArchiveStats
+}
+
+// Stats global stats regarding script execution
+type Stats struct {
+	StartTime  time.Time
+	EndTime    time.Time
+	TookTime   time.Duration
+	Output     *bytes.Buffer
+	Containers ContainersStats
+	BackupFile BackupFileStats
+	Archives   ArchivesStats
+}
+
+// NotificationData data to be passed to the notification templates
+type NotificationData struct {
+	Error  error
+	Config *Config
+	Stats  *Stats
+}
+
+type Config struct {
 	BackupSources              string        `split_words:"true" default:"/backup"`
 	BackupFilename             string        `split_words:"true" default:"backup-%Y-%m-%dT%H-%M-%S.tar.gz"`
 	BackupFilenameExpand       bool          `split_words:"true"`
@@ -148,15 +198,18 @@ var msgBackupFailed = "backup run failed"
 func newScript() (*script, error) {
 	stdOut, logBuffer := buffer(os.Stdout)
 	s := &script{
-		c: &config{},
+		c: &Config{},
 		logger: &logrus.Logger{
 			Out:       stdOut,
 			Formatter: new(logrus.TextFormatter),
 			Hooks:     make(logrus.LevelHooks),
 			Level:     logrus.InfoLevel,
 		},
-		start:  time.Now(),
-		output: logBuffer,
+		stats: &Stats{
+			StartTime: time.Now(),
+			Output:    logBuffer,
+			Archives:  ArchivesStats{},
+		},
 	}
 
 	if err := envconfig.Process("", s.c); err != nil {
@@ -169,7 +222,7 @@ func newScript() (*script, error) {
 		s.c.BackupLatestSymlink = os.ExpandEnv(s.c.BackupLatestSymlink)
 		s.c.BackupPruningPrefix = os.ExpandEnv(s.c.BackupPruningPrefix)
 	}
-	s.file = timeutil.Strftime(&s.start, s.file)
+	s.file = timeutil.Strftime(&s.stats.StartTime, s.file)
 
 	_, err := os.Stat("/var/run/docker.sock")
 	if !os.IsNotExist(err) {
@@ -275,7 +328,14 @@ func newScript() (*script, error) {
 		})
 	}
 
-	tmpl, err := template.ParseGlob("/etc/dockervolumebackup/notifications.d/*.*")
+	tmpl := template.New("")
+	tmpl.Funcs(template.FuncMap{
+		"formatTime": func(t time.Time) string {
+			return t.Format(time.RFC3339)
+		},
+	})
+
+	tmpl, err = tmpl.ParseGlob("/etc/dockervolumebackup/notifications.d/*.*")
 	if err != nil {
 		return nil, fmt.Errorf("unable to parse notifications templates: %w", err)
 	}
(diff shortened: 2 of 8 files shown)


cmd/backup/main.go
```diff
@@ -273,6 +275,12 @@ func newScript() (*script, error) {
 		})
 	}
 
+	tmpl, err := template.ParseGlob("/etc/volume-backup.d/*notifications.tmpl")
```
m90, line 278: Two questions here:
- should this maybe use `/etc/volume-backup/notifications.d` so it's more future proof (maybe some other feature wants to add something to `/etc/volume-backup` at some point
- should we just parse all files in this directory, no matter the naming? This `00foo.tmpl` is just a convention that is used often, but I think in theory most tools will accept everything.

cmd/backup/main.go
```diff
@@ -283,28 +291,43 @@ func (s *script) registerHook(level hookLevel, action func(err error) error) {
...
-// notifyFailure sends a notification about a failed backup run
-func (s *script) notifyFailure(err error) error {
-	body := fmt.Sprintf(
-		"Running docker-volume-backup failed with error: %s\n\nLog output of the failed run was:\n\n%s\n", err, s.output.String(),
-	)
-	title := fmt.Sprintf("Failure running docker-volume-backup at %s", s.start.Format(time.RFC3339))
+// notify sends a notification using the given title and body templates.
+// Automatically adds common params
+func (s *script) notify(titleTemplate string, bodyTemplate string, params map[string]string) error {
+	params["StartTime"] = s.start.Format(time.RFC3339)
```
m90, line 297: This is just discussing options: instead of using a `map[string]string` to hold the data that is passed to the templates, we could also define a struct type we then pass down. This could look for example like:

```
type NotificationData struct {
  StartTime string
  LogOuptut string
  Error error
  Config *config // this would make it really easy to access all config in the templates
}
```
That's probably a bit more verbose than using an ad-hoc map, but it might be easier to maintain as it implicitly documents what users can use in templates. Note to myself: this would also mean the definition of the `config` type becomes public API surface.

cmd/backup/main.go
```diff
@@ -283,28 +291,43 @@ func (s *script) registerHook(level hookLevel, action func(err error) error) {
...
+		return fmt.Errorf("notifyFailure: error executing %s template: %w", titleTemplate, err)
+	}
+	title := strings.Trim(titleBuf.String(), "\n")
+
+	bodyBuf := &bytes.Buffer{}
+	err = s.template.ExecuteTemplate(bodyBuf, bodyTemplate, params)
+	if err != nil {
+		return fmt.Errorf("notifyFailure: error executing %s template: %w", bodyTemplate, err)
+	}
+	body := strings.Trim(bodyBuf.String(), "\n")
```
m90, line 313: Only 99% sure, but I think Go templates can do this for us if you use '{{- define "foo" -}}`: https://pkg.go.dev/text/template#hdr-Actions

Dockerfile
```diff
@@ -18,6 +18,7 @@ RUN apk add --update ca-certificates
 COPY --from=builder /app/backup /usr/bin/backup
 
 COPY ./entrypoint.sh /root/
+COPY ./00notifications.tmpl /etc/volume-backup.d/
```
m90, line 21: This is very nit-picky, but I think I'd prefer if the file in the repo was called `notifications.tmpl` only and we prepend the `00` when copying:
```
COPY ./notifications.tmpl /etc/volume-backup.d/00defaults.tmpl
```

m90 (commented):
This is really good, thank you! I tested this and it works well for me.

I left some comments inline, let me know what you think.

cmd/backup/main.go
m90: If we want to be super consistent: the lockfile is currently named `"/var/lock/dockervolumebackup.lock"` so maybe this should be `/etc/dockervolumebackup/notifications.d/*.*`?

m90:
> but some things would have required too much refactor of the code to have this variables ready in the notifySuccess function. Any thoughts on this?

My initial idea about approaching this would be the following: instead of having `*script.start` and `*script.output` we introduce a type like this:

```go
type stats struct {
  Start time.Time
  Output *bytes.Buffer
  PrunedBackups int
  // whatever we want to add ...
}
```

Any method on `*script` could now write to these stats whenever needed and the `notify` functions could pass it down to the templates.

cmd/backup/main.go
m90: Just tested this, and it seems to work when used like this:

```
{{ define "body_failure" -}}
Running docker-volume-backup failed with error: {{ .Error }}

Log output of the failed run was:

{{ .LogOutput }}
{{- end }}
```

Dockerfile
MMauro94: No problem!

cmd/backup/main.go
MMauro94: Oops, I missed this feature. I'll remove the trim and change the default template then.

cmd/backup/main.go
m90: Another idea here would be to embed the default template in the binary itself and only read overrides from that dir:

```go
	var tmplErr error
	tmpl := template.New("notifications")
	tmpl, tmplErr = tmpl.Parse(defaultNotifications)
	if tmplErr != nil {
		return nil, fmt.Errorf("newScript: unable to parse default notifications templates: %w", err)
	}

	if _, err := os.Stat("/etc/dockervolumebackup/notifications.d"); err == nil {
		tmpl, tmplErr = tmpl.ParseGlob("/etc/dockervolumebackup/notifications.d/*.*")
		if tmplErr != nil {
			return nil, fmt.Errorf("newScript: unable to parse user defined notifications templates: %w", err)
		}
	}

	s.template = tmpl
```

The embed would happen at top level then:

```go
import _ "embed"

//go:embed notifications.tmpl
var defaultNotifications string
```

cmd/backup/main.go
MMauro94: I agree on creating a dedicated `struct` for this purpose.
Regarding passing the whole config, it's up to you if you want it to make public API. I'll wait for your decision :)

m90: I think if we start offering that feature, let's make it as useful as possible, and to me this also includes making the entire config avalilable. Considering the `ENV_VAR_NAMES` are public API anyways, i guess it's ok to keep the internal names locked as well. Let's do it.

MMauro94:
I agree, that's the approach I was going to suggest.

One thing: basically all the propertis in the `stats` struct will have to be passed down to the template, so I think it would be a good idea to add it to the notification data, like this:
```
type NotificationData struct {
  Error error
  Config *config
  Stats *stats
}
```

But what happens to the `*bytes.Buffer` output when being passed down to a template? Is it simply ignored? (In this case we'll need to keep the `LogOuptut string` in `NotificationData`)

cmd/backup/main.go
MMauro94: Perfect! One little question: I noticed that all the structs are lowercase in the project, but in your example `NotificationData` starts with an uppercase letter. Is this ok? What is the convention here?

m90: Docker for example also uses all uppercase in their templates i.e. `{{ .Container.Name }}`

MMauro94:
As you can see from the two commits I've implemented the discussed changes. 

From my testing everything works fine. I'll also share my success template:
```
{{ define "title_success" -}}
✅ Successfully ran backup {{ .Config.BackupStopContainerLabel }}
{{- end }}


{{ define "body_success" -}}
▶️ Start time: {{ .Stats.StartTime | formatTime }}
⏹️ End time: {{ .Stats.EndTime | formatTime }}
⌛ Took time: {{ .Stats.TookTime }}
🛑 Stopped containers: {{ .Stats.Containers.Stopped }}/{{ .Stats.Containers.All }} ({{ .Stats.Containers.StopErrors }} errors)
⚖️ Backup size: {{ .Stats.BackupFile.SizeBin }}
🗑️ Pruned backups: {{ .Stats.Archives.Local.Pruned }}/{{ .Stats.Archives.Local.Total }} ({{ .Stats.Archives.Local.PruneErrors }} errors)
{{- end }}
```

And the result:
image

Dockerfile
```diff
@@ -18,6 +18,7 @@ RUN apk add --update ca-certificates
 COPY --from=builder /app/backup /usr/bin/backup
 
 COPY ./entrypoint.sh /root/
+COPY notifications.tmpl /etc/dockervolumebackup/notifications.d/00defaults.tmpl
```
m90, line 21: Sorry for forgetting about this so long, but instead of copying over the file here, we could also have Go embed it at compile time, which would have the advantage that we can guarantee that any template in `notifications.d` will always override the default values (right now someone could mount `00adjustments` and would be surprised to see no effect)

This would work a little like this:

```go
	var tmplErr error
	tmpl := template.New("")
    // add funcs ...
	tmpl, tmplErr = tmpl.Parse(defaultNotifications)
	if tmplErr != nil {
		return nil, fmt.Errorf("newScript: unable to parse default notifications templates: %w", err)
	}

	if _, err := os.Stat("/etc/dockervolumebackup/notifications.d"); err == nil {
		tmpl, tmplErr = tmpl.ParseGlob("/etc/dockervolumebackup/notifications.d/*.*")
		if tmplErr != nil {
			return nil, fmt.Errorf("newScript: unable to parse user defined notifications templates: %w", err)
		}
	}

	s.template = tmpl
```

Embedding the file would need to happen at top level and works like this:

```go
import _ "embed"

//go:embed notifications.tmpl
var defaultNotifications string
```

What do you think?

cmd/backup/main.go
```diff
@@ -82,6 +83,8 @@ func main() {
 
 	s.must(s.encryptBackup())
 	s.must(s.copyBackup())
+	s.stats.EndTime = time.Now()
```
m90, line 86: Just curious: why did you decide to define "end" _before_ pruning happens?

cmd/backup/main.go
```diff
@@ -93,17 +96,66 @@ type script struct {
...
+	stats *Stats
 
-	c *config
+	c *Config
 }
 
-type config struct {
+// ContainersStats stats about the docker containers
+type ContainersStats struct {
+	All        int
```
m90, line 111: Not that it matters too much, but these could probably be `uint`s.

cmd/backup/main.go
```diff
@@ -93,17 +96,66 @@ type script struct {
...
+	WebDAV ArchiveStats
+	S3     ArchiveStats
+}
+
+// Stats global stats regarding script execution
+type Stats struct {
+	StartTime  time.Time
+	EndTime    time.Time
+	TookTime   time.Duration
+	Output     *bytes.Buffer
```
m90, line 145: Nit: maybe `LogOutput` would be a bit clearer, now that this has so many siblings.

cmd/backup/main.go
```diff
@@ -93,17 +96,66 @@ type script struct {
...
+
+// ArchivesStats stats about each possible archival location (Local, WebDAV, S3)
+type ArchivesStats struct {
+	Local  ArchiveStats
+	WebDAV ArchiveStats
+	S3     ArchiveStats
+}
+
+// Stats global stats regarding script execution
+type Stats struct {
```
m90, line 141: +1 I really like how you modeled these stats, amazing work :100:

cmd/backup/main.go
```diff
@@ -93,17 +96,66 @@ type script struct {
...
+	StartTime  time.Time
+	EndTime    time.Time
+	TookTime   time.Duration
+	Output     *bytes.Buffer
+	Containers ContainersStats
+	BackupFile BackupFileStats
+	Archives   ArchivesStats
```
m90, line 148: `Archives` as a name makes sense, but somehow collides with the "tar archive" terminology. I can't really come up with a better name right now though. The docs use the term `Storage`, but `Storages` sounds _very_ odd.

In case something pops up in my mind soon I will let you know, if not I guess this is just fine.

cmd/backup/main.go
```diff
@@ -93,17 +96,66 @@ type script struct {
...
+	Stopped    int
+	StopErrors int
+}
+
+// BackupFileStats stats about the created backup file
+type BackupFileStats struct {
+	Name      string
+	FullPath  string
+	SizeBytes int64
+	SizeBin   string
```
m90, line 122: Considering that you already added a `FuncMap` for the template I wonder if these calculations should rather happen on-demand in the template?

cmd/backup/main.go
```diff
@@ -856,6 +984,26 @@ func join(errs ...error) error {
 	return errors.New("[" + strings.Join(msgs, ", ") + "]")
 }
 
+// bytesToString converts an amount of bytes in a human-readable representation
+// the decimal parameter specifies if using powers of 10 (decimal) or powers of 2 (binary)
+func bytesToString(b int64, decimal bool) string {
```
m90, line 989: See the comment about how we could maybe make this a template function.

m90 (commented):
This is going to be brilliant. Thanks for putting so much thought into this, it's really appreciated.

I left a few comments inline once more, nothing really major though.

We'll also need docs for this I guess. Do you think a new section in `How to` would be enough considering we'll have to document the entire `NotificationData` type?

cmd/backup/main.go
```diff
@@ -273,6 +328,19 @@ func newScript() (*script, error) {
 		})
 	}
 
+	tmpl := template.New("")
+	tmpl.Funcs(template.FuncMap{
+		"formatTime": func(t time.Time) string {
```
m90, line 333: Is this currently used at all?

Dockerfile
MMauro94: I agree. That's neat that you can directly import plaintext files!

cmd/backup/main.go
MMauro94: Let's say that I have a big volume, like the Plex database, which consists of a lot of files, and is quite big (for example mine has 400k files and is 21 GiB). The "timeline" of actions, along with their duration, could look something like this:
```
20min:		tar the archive
5min:		encrypt the archive
1h:		upload to remote host
1h30min:	wait retention leeway period
10sec:		prune backups
10sec:		send notifications
```
The retention period babsically has to be the sum of previous operations, since the pruning algorithm checks for the last modified date on the files.
Now if someone asks me how much time it takes for the backup to run, I'd reply about 1 and a half hours, not 3 hours because of the delayed clean operation.


Now that I've written all this, it occured to me that there is a way simpler way of handling pruning: why not simply add to the deadline `BackupPruningLeeway` and get rid of the sleep. Wouldn't it achieve the same exact result? Or is there a particular reason for the sleep to be used?

The code would then simply be:
```go
deadline := time.Now().AddDate(0, 0, -int(s.c.BackupRetentionDays)).Add(s.c.BackupPruningLeeway)
```

cmd/backup/main.go
MMauro94: Agreed

cmd/backup/main.go
MMauro94: Agreed

cmd/backup/main.go
MMauro94: Thanks :)

cmd/backup/main.go
MMauro94: You're right. Also the word "storage" is mentioned a lot in the configuration reference, so it makes sense. I agree that `Storages` sounds a little odd, [but is apparently correct](https://www.wordhippo.com/what-is/the-plural-of/storage.html).

If you really don't like it, we could keep "storage", but avoid pluralizing and use something like `StorageTypes`, `StorageLocations`, `StorageList`, etc...

The only other somewhat similar word that I found is `Repository`, but I don't think it fits very well.

cmd/backup/main.go
MMauro94: What is? Are you referring to the `template.New("")` call?

cmd/backup/main.go
MMauro94: Completely agree!

cmd/backup/main.go
m90: Thanks for explaining, that makes sense.

> The retention period babsically has to be the sum of previous operations, since the pruning algorithm checks for the last modified date on the files.

If I recall correctly, it should be the fluctuation between slow and fast backup runs, not the entire time span. The sleep is there to prevent the following from happening:
- you have a very slow backup run (your system is busy or whatever) that takes 6 minutes
- 7 days later the next backup run takes the expected 5 minutes and looks for backups to prune
- the slow backup is not considered as it's still inside of the 7 day retention period

A leeway of 2 minutes should solve this, there's no need to use 5. Thishas been introduced ages ago https://github.com/offen/docker-volume-backup/commit/2892369677b042262a13c52c3b8a1547bce94ebc when this was still a shell script.

> it occured to me that there is a way simpler way of handling pruning: why not simply add to the deadline BackupPruningLeeway and get rid of the sleep.

You are totally correct. I think I just blindly ported this from the shell version (where it had to be done like this as you wouldn't have fine grained control over the deadline as easily) and never thought about this option.

Do you think it makes sense to change this in the context of this PR? In case it does not make your life easier, it might be easier to do it in a separate step for bookkeeping and such.

cmd/backup/main.go
m90: I guess `Storages` is probably the most consistent term to use.

cmd/backup/main.go
m90: Sorry for the confusion. I was referring to the `formatTime` function that does not seem to be used in the templates right now.

cmd/backup/main.go
MMauro94: > If I recall correctly, it should be the fluctuation between slow and fast backup runs, not the entire time span.

Of course, you're right! Note to self: think twice when working late at night.

> Do you think it makes sense to change this in the context of this PR? 

While not entirely on scope for the PR, it's a very easy fix and would be in its own commit, so I don't think it's a problem doing it now. Let me know if I should proceed.

cmd/backup/main.go
MMauro94: Yeah I agree. `Storages` it is then?

cmd/backup/main.go
MMauro94: I was 100% sure I added it in the default template. I probably forgot, but it is needed because of how instances of `Time` are converted to string by default (see here https://go.dev/play/p/s10JwKYEVzX)

In the upcoming changes it will surely be used in the default template.

MMauro94:
> We'll also need docs for this I guess. Do you think a new section in `How to` would be enough considering we'll have to document the entire `NotificationData` type?

I think the best way is to add a section in the `How to` that explains the feature, but leave the full documentation of `NotificationData` and the funcs in a separate file, that will be of course be linked from the how to section.

What do you think?

cmd/backup/main.go
m90: Yeah, let's include. It makes a lot of sense and will save quite some time for some people.

cmd/backup/main.go
m90: Ok, brilliant.

MMauro94:
I have took care of all the issues you mentioned. Also, I have drafted the initial documentation for the notification templates. 

Let me know what do you think :)

m90:
> Is it possible that the reason is that I've removed the sleep in the pruning, thus not leaving docker enough time to start up the containers before the check is made?

Considering `docker ps` prints the following for the failing test (notice the `Status: Created`):

```
CONTAINER ID        IMAGE                                      COMMAND                  CREATED             STATUS                     PORTS               NAMES
6f9cddc5ca40        postgres:12.2-alpine                       "docker-entrypoint.s…"   2 seconds ago       Up Less than a second      5432/tcp            test_stack_pg.1.v5xo597bkr5p87xzhndkw92lq
4d9334b0ecb0        offen/offen:latest                         "/sbin/tini -- offen"    2 seconds ago       Created                                        test_stack_offen.2.i9b6yneky00vz1it2ep0gxts3
15f16d539f49        postgres:12.2-alpine                       "docker-entrypoint.s…"   24 seconds ago      Exited (0) 4 seconds ago                       test_stack_pg.1.wk7mfmsoefwtiubaxfbb5dsew
20148f0030d4        offen/docker-volume-backup:canary          "/root/entrypoint.sh"    29 seconds ago      Up 25 seconds                                  test_stack_backup.1.bd4349hxlf0w3ap0tqxyr5gv0
9e586796d160        minio/minio:RELEASE.2020-08-04T23-10-51Z   "/bin/ash -c 'mkdir …"   30 seconds ago      Up 26 seconds              9000/tcp            test_stack_minio.1.y322im0htf6yf96rucucung8b
21e9a0de1407        offen/offen:latest                         "/sbin/tini -- offen"    30 seconds ago      Exited (0) 3 seconds ago                       test_stack_offen.2.79fst9vq9i0gmztm8r9vp4psd
50ef1d4b3876        offen/offen:latest                         "/sbin/tini -- offen"    30 seconds ago      Exited (0) 3 seconds ago                       test_stack_offen.1.r1ibf6j4s39yi0jgs815z8zqc
```

I think it's very likely this is a race condition created by removing the `sleep`. I think it'd be ok to add a short sleep before this check runs: https://github.com/offen/docker-volume-backup/blob/3bb99a711754ba53f549c70c8d0914b6007e329d/test/swarm/run.sh#L26-L30

I'll have a close look at the PR itself later today.

MMauro94:
> I think it'd be ok to add a short sleep before this check runs

Agreed. Do you think 5 seconds is enough?

Looking at the test directory, I think the right place is just before these three lines:
* https://github.com/offen/docker-volume-backup/blob/3bb99a711754ba53f549c70c8d0914b6007e329d/test/cli/run.sh#L53
* https://github.com/offen/docker-volume-backup/blob/3bb99a711754ba53f549c70c8d0914b6007e329d/test/compose/run.sh#L32
* https://github.com/offen/docker-volume-backup/blob/3bb99a711754ba53f549c70c8d0914b6007e329d/test/swarm/run.sh#L26

Is it correct?

m90:
I think we only need to add it for the `swarm` test:
- the `cli` test doesn't stop containers on purpose, so there's no restart to wait for
- considering the `compose` test does some gpg decryption before this gives us some implicit sleep that should be enough to cater for this setup

---

That being said: it could be argued that the compose test should check that containers are up _before_ it looks at the artifacts, which would probably also require the sleep to be introduced.

Long story short: it would be great if you could add the sleep for the swarm test and I'll see if I can clean up and also document the tests a little better soon (I'd also like to add a test for the notifications feature at some point, but this will be a little tricky).

docs/NOTIFICATION-TEMPLATES.md
```diff
@@ -0,0 +1,38 @@
+# Notification templates reference
+
+In order to customize title and body of notifications you'll have to write a [go templates](https://pkg.go.dev/text/template) and mount it inside the `/etc/dockervolumebackup/notifications.d/` directory.
```
m90, line 3:
```suggestion
In order to customize title and body of notifications you'll have to write a [go template](https://pkg.go.dev/text/template) and mount it inside the `/etc/dockervolumebackup/notifications.d/` directory.
```

docs/NOTIFICATION-TEMPLATES.md
```diff
@@ -0,0 +1,38 @@
...
+
+Data and functions will be passed to it, this page documents it fully.
```
m90, line 5:
```suggestion
Configuration, data about the backup run and helper functions will be passed to this template, this page documents them fully.
```

Dockerfile
```diff
@@ -6,6 +6,7 @@ FROM golang:1.17-alpine as builder
 WORKDIR /app
 COPY go.mod go.sum ./
 RUN go mod download
+COPY cmd/backup/notifications.tmpl ./cmd/backup/notifications.tmpl
```
m90, line 9: I think we could also just do

```
COPY cmd/backup ./cmd/backup/
```

now. Which will also create a good incentive for me to start splitting this "little file" that has grown to be 1000+LOC by now into chunks at some point :)

m90 (commented):
This turned out really well. I left some minor suggestions, feel free to adjust or disregard these as you wish. If we can also get CI green again I would say this one is ready to go :tada:

MMauro94:
It seems to be fixed with the sleep. I've also implemented the other changes.

m90 (approved):

m90:
This is now released in [`v2.11.0`](https://github.com/offen/docker-volume-backup/releases/tag/v2.11.0). Thanks a lot for this top-notch contribution :tophat:

MMauro94:
Thanks to you for the project and your quick to the point assistance 🔥

PR #66, opened 2022-02-14
Fixed TookTime

MMauro94:
Hi, it's me again! I noticed that my backups strangely always took 0 seconds, so I investigated and found this bug.


m90:
I noticed that too when I added the test cases but thought it was me... Thanks for catching and fixing. I will merge and release once I'm back at a keyboard later today.

m90:
This is released in `v2.11.2` :rocket:

PR #76, opened 2022-02-22
snapshot-style restore example

kzshantonu:


m90 (approved):
Thank you for this great addition :100:

PR #89, opened 2022-04-12
Fixed wrong env name for S3 bucket name in README

simboel:
The README was using `AWS_BUCKET_NAME` instead of `AWS_S3_BUCKET_NAME` in the recipes. 
This resulted in no data being uploaded to S3.


m90:
This is a very good find, thanks for fixing!

Issue #65, opened 2022-02-13
Support scp backed storage


m90:
Given the current architecture of this script, it would be relatively easy to add support for storing backups on remote systems that are accessed by SSH / `scp`.

Open questions:
- how does pruning of old backups work in this scenario? `scp` itself does only copy. Is there another suitable Go library for syncing files over the wire?
- how does authentication work? Do users mount authorized SSH keys into the container? Are there any security issues I am not aware of?

rpatel3001:
a better idea might be supporting rclone and having users mount an rclone.conf file. instant support for a ton of backends

rpatel3001:
Nope, go was too much trouble. A couple hiccups but I hooked up my shell script that syncs my latest backup using rclone to be triggered once a backup job has run. 

Add to the backup container itself:
```
    labels:
      - docker-volume-backup.exec-post=/bin/sh -c '/sync/setup.sh'
    volumes:
      - ./config/sync_backup:/sync:ro
```

And the contents of `./config/sync_backup`:
32 bit `rclone` executable
32 bit `shoutrrr` executable
`rclone.conf`, generated by running `rclone config --config ./config/sync_backup/rclone.conf`
any other files required by rclone (key files for remotes)

`setup.sh`:
```
#!/bin/sh
apk add --update openssh-client at
atd
# need to run the sync after the backups have been rotated and the symlink updated
echo /sync/sync_backup.sh | at now +10 minutes
```

`sync_backup.sh`:
```
#!/bin/sh

# I'm using a single URL, this might not work with multiple
CMD="send -u $NOTIFICATION_URLS"

# tell me how big each of my volumes, mounted as folders under /backup, takes up
/sync/shoutrrr $CMD -t "Backed up volume sizes" -m "`du -sm /backup/* | sort -rg | sed 's#/backup/##' | sed 's#\([0-9]\+\)#\1M#'`"

counter=0
# measure how long the upload takes
START_TIME=`date +%s`
echo "Uploading to Google Drive"
until
   /sync/rclone copy --config /sync/rclone.conf -vL /archive/backup-latest.tar.gz gdrive:
do
   sleep 1
   [ "$counter" -eq "$max_retry" ] && echo "Failed!" && break
   echo "Trying again. Try #$counter"
   counter=$((counter+1))
   /sync/shoutrrr $CMD -t "Backup sync" -m "google drive upload failed on iteration $counter/$max_retry"
done
END_TIME=`date +%s`
[...]

PR #107, opened 2022-06-15
SSH Backup Storage Support

Riscue:
SSH and SFtp feature added.

Copying and pruning working exactly same as other storage services.

Readme updated.

Tests updated.


Riscue:
#65

m90:
Yeah it seems as if Public Key authentication is supported https://pkg.go.dev/golang.org/x/crypto/ssh#AuthMethod

Maybe it could work like this:
- when `SSH_HOST_NAME` is given but not `SSH_PASSWORD`, we try using key authentication, assuming a keypair is mounted in `/root/.ssh` (maybe we need to make the name of the key configurable too, not sure right now)
- in case `SSH_PASSWORD` is given, password authentication is used

What do you think?

README.md
```diff
@@ -245,6 +246,29 @@ You can populate below template according to your requirements and use it as you
 
 # WEBDAV_URL_INSECURE="true"
 
+# You can also backup files to any SSH server:
+
+# The URL of the remote SSH server
+
+# SSH_HOST_NAME="server.local"
```
m90, line 253: Hostname usually is spelled as a single word, so I'd probably make this `SSH_HOSTNAME`, also `SSHHostname`

Riscue:
> Yeah it seems as if Public Key authentication is supported https://pkg.go.dev/golang.org/x/crypto/ssh#AuthMethod
> 
> Maybe it could work like this:
> 
> * when `SSH_HOST_NAME` is given but not `SSH_PASSWORD`, we try using key authentication, assuming a keypair is mounted in `/root/.ssh` (maybe we need to make the name of the key configurable too, not sure right now)
> * in case `SSH_PASSWORD` is given, password authentication is used
> 
> What do you think?

I am newbie at golang. I tried publickey auth but i was facing some issues couldn't resolve. I am not allowing password auth either. It would be nice. I will also try to make the corrections you mentioned in the comments. Thanks for review.

README.md
Riscue: I used naming as in the `.ssh/config` file.

```
Host github.com
        HostName github.com
        User git
        IdentityFile ~/.ssh/github.com/id_rsa
```

m90:
> I am newbie at golang.

It looks pretty solid already :) feel free to ask for help if you run into any problems and I'll see if I can help out.

README.md
m90: Good point, I never noticed it's spelled like this in there. Then let's keep it `HOST_NAME` I guess.

cmd/backup/config.go
```diff
@@ -45,6 +45,13 @@ type Config struct {
 	WebdavPath                 string        `split_words:"true" default:"/"`
 	WebdavUsername             string        `split_words:"true"`
 	WebdavPassword             string        `split_words:"true"`
+	SSHHostName                string        `split_words:"true"`
+	SSHPort                    string        `split_words:"true" default:"22"`
+	SSHUser                    string        `split_words:"true"`
+	SSHPassword                string        `split_words:"true"`
+	SSHIdentityFile            string        `split_words:"true" default:"/root/.ssh/id_rsa"`
```
m90, line 52: I wonder if we can skip this option and mandate mounting the key file at `/root/.ssh/id`. This could possibly only conflict with someone extending the image, but used as-is I cannot think of any scenario where this would become a problem right now. Any opinion on this?

cmd/backup/script.go
```diff
@@ -159,6 +164,45 @@ func newScript() (*script, error) {
 		}
 	}
 
+	if s.c.SSHHostName != "" {
+		var authMethods []ssh.AuthMethod
```
m90, line 168: Just for my own understanding, what is the precedence in case you specify multiple methods? Does it go right to left or left to right?

cmd/backup/script.go
```diff
@@ -159,6 +164,45 @@ func newScript() (*script, error) {
...
+				} else {
+					signer, err = ssh.ParsePrivateKey(key)
+				}
+				if err == nil {
+					authMethods = append(authMethods, ssh.PublicKeys(signer))
+				}
+			}
+		}
+
+		sshClient, err := ssh.Dial("tcp", fmt.Sprintf("%v:%v", s.c.SSHHostName, s.c.SSHPort), &ssh.ClientConfig{User: s.c.SSHUser, Auth: authMethods, HostKeyCallback: ssh.InsecureIgnoreHostKey()})
```
m90, line 189: Two nitpicks:
- can we break this into multiple lines, it's very long
- can we use `%s` verbs in the `fmt.Sprintf` calls? We do know we are passing strings and don't need magic reflection under the hood to figure out how to stringify these values

cmd/backup/script.go
```diff
@@ -159,6 +164,45 @@ func newScript() (*script, error) {
...
+		s.sshClient = sshClient
+		if err != nil {
+			return nil, fmt.Errorf("newScript: ssh new client: %w", err)
```
m90, line 192:
```suggestion
			return nil, fmt.Errorf("newScript: error creating ssh client: %w", err)
```

cmd/backup/script.go
```diff
@@ -159,6 +164,45 @@ func newScript() (*script, error) {
...
+		}
+		_, _, err = s.sshClient.SendRequest("keepalive", false, nil)
+		if err != nil {
+			return nil, err
+		}
+
+		sftpClient, err := sftp.NewClient(sshClient)
+		s.sftpClient = sftpClient
+		if err != nil {
+			return nil, fmt.Errorf("newScript: sftp new client: %w", err)
```
m90, line 202:
```suggestion
			return nil, fmt.Errorf("newScript: error creating sftp client: %w", err)
```

cmd/backup/script.go
```diff
@@ -512,6 +556,52 @@ func (s *script) copyBackup() error {
...
+		}
+		defer source.Close()
+
+		destination, err := s.sftpClient.Create(filepath.Join(s.c.SSHRemotePath, name))
+		if err != nil {
+			return fmt.Errorf("copyBackup: error creating file on SSH storage: %w", err)
+		}
+		defer destination.Close()
+
+		chunk := make([]byte, 1000000)
```
m90, line 572: Just being curious: where does this number come from?

cmd/backup/script.go
```diff
@@ -159,6 +164,45 @@ func newScript() (*script, error) {
...
+	if s.c.SSHHostName != "" {
+		var authMethods []ssh.AuthMethod
+
+		if s.c.SSHPassword != "" {
+			authMethods = append(authMethods, ssh.Password(s.c.SSHPassword))
+		}
+
+		if _, err := os.Stat(s.c.SSHIdentityFile); err == nil {
+			key, err := ioutil.ReadFile(s.c.SSHIdentityFile)
+			if err == nil {
```
m90, line 176: I wonder if it's ok to just continue in case there is an error here? If we reach this point and `err` is not `nil` it means an identity file exists, but there was an error reading it (e.g. permissions or whatever). I would expect the script to fail in such cases instead of trying to continue without the key file. Any reason for this behavior I am missing?

test/compose/run.sh
```diff
@@ -2,9 +2,10 @@
 
 set -e
 
-cd $(dirname $0)
+cd "$(dirname "$0")"
 
 mkdir -p local
+ssh-keygen -t rsa -m pem -b 4096 -N "" -f id_rsa -C "docker-volume-backup@local"
```
m90, line 8: +1 for going the extra mile and testing the feature with a key file

m90 (commented):
I left another round of questions, but again there are no major issues. Thanks!

cmd/backup/config.go
Riscue: If in any circumstances this will be a problem or conflict with someone, the default path can be changed with this environment. I can't think of any case right now

m90: Should we make the default `/root/.ssh/id` then? It'd feel weird to map a non-RSA key (ED25519 for example) to a file called like this. people might even thing it HAS to be a RSA key?

cmd/backup/script.go
Riscue: I have set up a ssh server with password auth and publickey auth simultaneously. And i debugged app with all cases.

Password | Publickey | Auth
-------- | ---------- | ----------
| :heavy_check_mark: | - | :heavy_check_mark: |
| - | :heavy_check_mark: | :heavy_check_mark: |
| :heavy_check_mark: | ✗ | :heavy_check_mark: |
| ✗ | :heavy_check_mark: | :heavy_check_mark: |
| :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: |
| - | - | ✗ |
| ✗ | - | ✗ |
| - | ✗ | ✗ |
| ✗ | ✗ | ✗ |


When both password and private key wrong error log is something like that:
`panic: newScript: ssh new client: ssh: handshake failed: ssh: unable to authenticate, attempted methods [none password publickey], no supported methods remain`

If we change the password auth method and private key method order in the code. err message changing to:

`panic: newScript: ssh new client: ssh: handshake failed: ssh: unable to authenticate, attempted methods [none publickey password], no supported methods remain`

There is a for loop  that in client_auth.go:64 `for _, a := range config.Auth`

Ssh package firstly tries `none auth method`. Next tries the methods provided order. That means which method provided first it uses according to that order.

cmd/backup/config.go
m90: The devil is in the details. `man ssh` tells me:

```
     -i identity_file
             Selects a file from which the identity (private key) for public key authentication is read.  The default is ~/.ssh/id_dsa, ~/.ssh/id_ecdsa, ~/.ssh/id_ecdsa_sk, ~/.ssh/id_ed25519,
             ~/.ssh/id_ed25519_sk and ~/.ssh/id_rsa.  Identity files may also be specified on a per-host basis in the configuration file.  It is possible to have multiple -i options (and multiple
             identities specified in configuration files).  If no certificates have been explicitly specified by the CertificateFile directive, ssh will also try to load certificate information
             from the filename obtained by appending -cert.pub to identity filenames.
```

Maybe we can leave `id_rsa` as the default, but add a note to the documentation about how non-RSA key algos are supported as well?

cmd/backup/script.go
m90: Sounds perfect, thanks for the thorough explanation.

README.md
```diff
@@ -245,6 +246,39 @@ You can populate below template according to your requirements and use it as you
...
+
+# The private key path in container for SSH server
+# Default value: /root/.ssh/id_rsa
+# If file is mounted to /root/.ssh/id_rsa path it will be used.
+
+# SSH_IDENTITY_FILE="/root/.ssh/id_rsa"
+
+# The passphrase for the identity file
+
+# SSH_PASSPHRASE="pass"
```
m90, line 280: Just an idea: to reduce possible mixups of password and passphrase, how about naming this `SSH_IDENTITY_PASSPHRASE` or `SSH_KEY_PASSPHRASE`?

Riscue: Changed to SSH_IDENTITY_PASSPHRASE and test updated with passphrase

cmd/backup/config.go
Riscue: You are right and i didnt read that. I changed to .ssh/id and updated documentation.

m90:
Brilliant, thank you very much!

cmd/backup/config.go
m90: Ok, then I'll change it myself real quick before cutting the release, sorry for the back and forth.

Riscue:
Thank you! For your helps and this awesome solution. I needed some backup mechanism on my setup.

m90:
This is now released in [v2.18.0](https://github.com/offen/docker-volume-backup/releases/tag/v2.18.0) :tada:

Issue #116, opened 2022-07-04
Add custom headers for S3


rwanito:
* **I'm submitting a ...**
  - [ ] bug report
  - [x] feature request
  - [ ] support request

* **What is the current behavior?**

I would like to add the headers support for S3 (ex: cold storage).
For example : `x-amz-storage-class` to GLACIER or STANDARD.
I'm not a go dev, so I prefer to let someone stronger than me here :p !

Thanks !

m90:
This image is using the MinIO Golang SDK and it seems it does not allow setting headers like you propose but instead defines its own `PutObjectOptions`: https://github.com/minio/docs/blob/master/source/sdk/go/API.md#putobjectctx-contextcontext-bucketname-objectname-string-reader-ioreader-objectsize-int64opts-putobjectoptions-info-uploadinfo-err-error

I wouldn't be opposed to someone adding more of these options if it makes sense though. What would need to happen is:
- add the option to the `config` struct, preferrably with an `AWS_S3_` prefix: https://github.com/offen/docker-volume-backup/blob/c2a8cc92fcfc05e974df4656f024161cdd464c40/cmd/backup/config.go#L14-L58
- pass the value to the Minio Client here: https://github.com/offen/docker-volume-backup/blob/c2a8cc92fcfc05e974df4656f024161cdd464c40/cmd/backup/script.go#L548-L555
- add docs

---

> I'm not a go dev, so I prefer to let someone stronger than me here :p !

This repo has seen great feature contributions by first timers already, so don't be hesitant to try. Go isn't really that hard to pick up.

m90:
Let me know if you need any help or input with that.

PR #119, opened 2022-07-06
feat: Add storage class header

rwanito:
Closes issue #116

### DIFF
--- a/cmd/backup/config.go
+++ b/cmd/backup/config.go
@@ -28,6 +28,7 @@ type Config struct {
 	AwsEndpoint                string        `split_words:"true" default:"s3.amazonaws.com"`
 	AwsEndpointProto           string        `split_words:"true" default:"https"`
 	AwsEndpointInsecure        bool          `split_words:"true"`
+	AwsStorageClass            string        `split_words:"true" default:"STANDARD"`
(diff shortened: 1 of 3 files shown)


cmd/backup/config.go
```diff
@@ -28,6 +28,7 @@ type Config struct {
 	AwsEndpoint                string        `split_words:"true" default:"s3.amazonaws.com"`
 	AwsEndpointProto           string        `split_words:"true" default:"https"`
 	AwsEndpointInsecure        bool          `split_words:"true"`
+	AwsStorageClass            string        `split_words:"true" default:"STANDARD"`
```
m90, line 31: If I understand the Minio Docs correctly, the default value is the empty string (this is what is used right now anyways), so I think we could get away without the default value here (although the behavior is likely the exact same).

m90 (commented):
Thanks for working on this. I left one minor comment inline but in general this is good to go. Could you also add the new option to the configuration reference in the README?

rwanito:
Okay, I changed the default value and the README !
Tell me when you think it is good !

Later, when the bump will be done in docker, I'll give an example to set the backing up for Scaleway provider.

Issue #113, opened 2022-06-23
Define abstract interface for storage backends


m90:
@m90 Great! Quick question: there is quite a bit of similar code for the different types of storages. Would you be open to generalize the code a bit? In a OOP language I would create a common abstract class with with abstract functions like `listFiles`, `uploadFile`, `removeFile`, and so on. Then it's simply a matter of implementing each storage in its own class. This would allow, for instance, to have a single `purge` function. I'm guessing the go equivalent would be something similar to [this](https://stackoverflow.com/a/47263226).

I know that it's a bit out of scope for this issue, but I'd like to hear your thoughts on this. Also, there to consider the possibility that a single storage jack-of-all-trades like rclone will be the way to go in the future (as per #65); in this case abstracting the current storages like this will probably be a waste of time.

_Originally posted by @MMauro94 in https://github.com/offen/docker-volume-backup/issues/94#issuecomment-1123224662_

Issue #103, opened 2022-06-02
[Feature Request] Support backup targets like Box / Dropbox


lonix1:
* **I'm submitting a ...**
  - [ ] bug report
  - [x] feature request
  - [ ] support request

* **What is the current behavior?**
Backups can be made to a variety of S3-like targets, but not storage services like Dropbox, Box, etc.

* **What is the motivation / use case for changing the behavior?**
These are very common in business environments. 

I'm not a go dev, but maybe someone else knows how to do this... Thanks!

m90:
If anyone wants to add this, there seems to be an official (published by Dropbox) "unofficial" Dropbox SDK for Go: https://github.com/dropbox/dropbox-sdk-go-unofficial which can do all we need to do here.

lonix1:
Nice find.

For Box: has many dev tools [and SDKs](https://developer.box.com/sdks-and-tools/#official-sdks-clis) but not an official Go SDK; however it does have [a CLI](https://github.com/box/boxcli) which (I assume) could be wrapped in go code.

Box is nice because their free tier is a massive 10GiB vs DropBox's 2GiB. Both are good though.

m90:
Dropbox is now supported as of [v2.30.0](https://github.com/offen/docker-volume-backup/releases/tag/v2.30.0)

MaxJa4:
Haven't heard of Box before. Is it widely used in some countries or am I the only one who never heard of them before?

On another note, if we include Dropbox, perhaps Google Drive make sense as an option too. There is an official Go SDK for it and it also uses OAuth2 like Dropbox.

m90:
> The latter I'm afraid :-) Box is gigantic and in the same league as Dropbox - and like I said above very common in business environments.

Gotta be honest, I also never came across Box outside of this issue.

> On another note, if we include Dropbox, perhaps Google Drive make sense as an option too. 

Google Drive is probably heavily used, which is good, but I am a bit worried about API stability here as Google is pretty notorious for just changing things when they feel like it. It's probably not possible to connect to Google Drive using some open standard, e.g. WebDAV, SSH, S3? Also: as we've just been bitten by this: is it possible to run a mock Google Drive in tests instead of having to resort to a live service?

MaxJa4:
[...]

Issue #101, opened 2022-05-09
Azure Storage Support


DevSecNinja:
* **I'm submitting a ...**
  - [ ] bug report
  - [x] feature request
  - [ ] support request

* **What is the expected behavior?**

Enabling docker-volume-backup to backup to Azure Storage.

* **What is the motivation / use case for changing the behavior?**

Azure Storage is a great solution (especially with cold storage) to backup Docker volumes to. I don't have any experience with Go but based on the AWS format I might be able to implement support for Azure Storage. Additionally, lots of professionals working with Microsoft technology have a subscription from work that they would like to use. (e.g. via a Visual Studio subscription)

SDK Docs: https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/storage/azblob#pkg-overview

Thanks for this awesome tool!

m90:
This would be a great addition indeed.

One question upfront: is it possible to spin up a Azure Blob Storage compatible tool in a Docker container so it can be used in tests for this image? I would like to avoid to a. create an Azure account just to be able to support this feature and b. run tests against remote infrastructure. Ideally, this could work like with S3 where the tests run a MinIO server or WebDAV where an instance of `bytemark/webdav` is running.

DevSecNinja:
> This would be a great addition indeed.
> 
> 
> 
> One question upfront: is it possible to spin up a Azure Blob Storage compatible tool in a Docker container so it can be used in tests (and development) for this image? I would like to avoid to a. create an Azure account just to be able to support this feature and b. run tests against remote infrastructure. Ideally, this could work like with S3 where the tests run a MinIO server or WebDAV where an instance of `bytemark/webdav` is running.

Absolutely and totally understandable! It can be done with Azurite: https://docs.microsoft.com/en-us/azure/storage/common/storage-use-azurite

DevSecNinja:
> Sounds good. If you or someone else wants to pick this up right away, I am happy to help with getting PRs merged (it should be relatively straight forward). Otherwise I'll keep it on the roadmap so we can get this supported sooner or later. Thanks for the suggestion.

Awesome! I must say that I'm also a part time CS grad student besides work, so I can't make any commitments. But it's high on my list to pick up when I have some spare time. Thanks!

m90:
> so I can't make any commitments

No need to worry here.

m90:
[...]

Issue #94, opened 2022-04-19
Flag to avoid backing up long-stopped containers


MMauro94:
It would be nice to be able to add a flag (either in the config or in the container labels) to avoid making a backup if the container is stopped AND it was last stopped before the latest backup.

My use case is this: I have a game server container that I start on demand. There may be periods on which it'll be used almost every day and periods of months when it will be permanently stopped. I would like to run backups only when there are actual changes to the state of the container. When used in conjunction with backup pruning this also avoids having N backups of the same exact thing and allows to keep the last N states.

By running `docker inspect` I found that there are the `StartedAt` and `FinishedAt` properties, so it should be relatively easy to pull this info up and compare it with the last backup date.

m90:
This is an interesting proposal, however I am not sure if it's possible to do this in a configuration friendly way because of the following: right now, a `docker-volume-backup` container does not really know how to map a volume that is mounted for backup to another container running on the host. Starting / stopping containers during backup happens using labels, but this still does not create any kind of logical connection to a volume that is being backed up.

So if we were to implement such a feature, we'd need to come up with an API that allows linking a volume and a service definition. Right now, I have a hard time coming up with something that would work here, but I'll play with it in the back of my head for a little while. Do you have a good idea how such an API could look like?

---

Looking at your underlying use case, another solution could be: introduce an option like `CONSIDER_STALE_AFTER` (a value would be something like `168h`), and in case no file in the volume has changed after the deadline defined by the option, backup is skipped. What do you think about this approach?

MMauro94:
You're right, as I am using a single specific `docker-volume-backup` container for this case so it didn't occur to me the problem about mapping volumes.

An option could be to make this a global config instead of a label in a container and apply the rule to all matching containers. The condition to avoid making a backup would then be: `all containers are stopped AND MAX FinishedAt < lastest_backup_date`.

---

In my case `CONSIDER_STALE_AFTER` would definetly work, but I have two questions:
1. Why define a fixed amount of time instead of simply checking `last_changed_file_date < lastest_backup_date`?
2. It doesn't apply to my case, but there may be a significant performance hit in directories with a lot of small files.

m90:
I think I prefer the second approach in the variation you describe. It's just very easy to understand and would cover a lot of cases. The option could be called something like `SKIP_IF_UNCHANGED`.

> Why define a fixed amount of time instead of simply checking last_changed_file_date < lastest_backup_date

[...]

PR #135, opened 2022-07-21
Added abstract helper interface for all storage backends

MaxJa4:
Implements #113 

I created an abstract helper interface and implemented it for all storage backends (Minio/S3, WebDav, SSH, local).
The obvious candidates for similar operations were copyArchive and pruneBackups as well as the initialization of the different clients.
Also doPrune has been moved to the abstract helper.

As I'm no Go expert, feel free to adjust these changes to your liking or suggest optimizations.
Maybe the helpers could be moved in their own module - idk how well that would work with Go though while avoiding import cycles since the helpers need "script" and the script.go needs the helpers...


m90:
Thanks for working on this. As it's a pretty big changeset and also a really important step forward, review might take a while and I just wanted to let you know that it's much appreciated and I'm looking forward to get this merged once it's ready. Thanks :tophat:

MaxJa4:
No worries! It's non-functional so there is no hurry.
Since it touches almost the entire application, it also helped getting to know the code for other contributions :)

MaxJa4:
Maybe having the helpers in a wrapper would be beneficial. One would only have to call helpers.prune() (as an example).
Having a list of helpers could work too, just calling helpers[i].prune() (as an example) in a loop.

MaxJa4:
Just saw your comment in #94, didn't know you already implemented something like that on a different branch. I'll have a look, maybe it can be combined or extended :)

m90:
> didn't know you already implemented something like that on a different branch

I wasn't super happy with that (otherwise I would have merged it), so feel free to stray away from that path when it makes sense.

MaxJa4:
To do:
[ ] Add collection of storage backends and unify them in script.go (full transparency if possible)
[ ] Polish/Cleanup
[ ] Summarize changes here

MaxJa4:
Changes:
- Moved files with only structs (config/stats) to sub-module "types" so it can be accessed by all other modules
- Moved util to sub-module "utilites" so it can be accessed by all other modules
- Created abstract interface / struct for storages (Storage/GenericStorage)
- Applied pattern to all implemented storage backends (especially regarding the methods init, copy and prune)
- Unified storage backends into a single StoragePool
- Script only has to call StoragePool once per action to work as before -> transparent, easy to use and extendable

All changes are just architectural and structural. I tried to not change any functional flows.
IMO it is a lot nicer now, easy to extend with more backends and intuitive to understand.

Feel free to have a look as you find time @m90 and share your thoughts - no hurry here.
After this is done, I could tackle #103 and/or #101 since it would be quite a bit easer to implement after this PR.

cmd/backup/hooks.go
```diff
@@ -6,6 +6,8 @@ package main
 import (
 	"fmt"
 	"sort"
+
+	u "github.com/offen/docker-volume-backup/cmd/backup/utilities"
```
m90, line 10: I'm not sure yet what to think about splitting this into multiple packages.

Go has very strong opinions on how to do this and there are also some conventions around "importability" that we should be following if we wanted to do this.

## Package location

Right now, the newly introduced packages are in `cmd` which looks as if they contain a runnable program when they don't. The proper location for such a "utilities" package would be `github.com/offen/docker-volume-backup/utilities` (or even `github.com/offen/utilities`). Doing this would create a new problem though: the package is now importable by any other Go project, which I don't really want to happen (API stability and such). The Go way of doing this would be making these packages "internal" which prevents foreign imports from happening. The proper location for this package would therefore probably be `github.com/offen/docker-volume-backup/internal/utilities`.

See this article: https://dave.cheney.net/2019/10/06/use-internal-packages-to-reduce-your-public-api-surface.

## Package contents

I'm not sure if the split into packages like `types` and `util` makes much sense here, as there is only a single consumer, and it makes the code more verbose. I could see the need for a `storage` package (which the also contains the types), but I'm not so sure about `types` and `util`. "If in doubt, keep it dumb".

Some extra context:
- https://github.com/golang/go/wiki/CodeReviewComments#package-names
- https://go.dev/blog/package-names

---

I'm sure you spent a lot of time and thought on this. What's your opinion here?

cmd/backup/script.go
```diff
@@ -26,56 +26,48 @@ import (
...
-	webdavClient *gowebdav.Client
-	sshClient    *ssh.Client
-	sftpClient   *sftp.Client
-	logger       *logrus.Logger
-	sender       *router.ServiceRouter
-	template     *template.Template
-	hooks        []hook
-	hookLevel    hookLevel
+	cli       *client.Client
+	providers *strg.StoragePool
```
m90, line 38: Would it make sense to use the name `pool` or `storagePool` when it contains a `StoragePool` instance? `providers` doesn't really map to the type for me.

cmd/backup/script.go
```diff
@@ -107,112 +99,9 @@ func newScript() (*script, error) {
 		s.cli = cli
 	}
 
-	if s.c.AwsS3BucketName != "" {
```
m90, line 110: I would prefer if we could keep all logic that conditionally creates a storage client or not based on the configuration in `func newScript`.

There's a comment

```
// All reading from env vars or other configuration sources is expected to happen
// in this method.
```

and I think it still makes sense to do it like this.

Each block could then add the storage to the pool or not.

cmd/backup/storages/local.go
```diff
@@ -0,0 +1,118 @@
...
+	"path"
+	"path/filepath"
+	"time"
+
+	t "github.com/offen/docker-volume-backup/cmd/backup/types"
+	u "github.com/offen/docker-volume-backup/cmd/backup/utilities"
+	"github.com/sirupsen/logrus"
+)
+
+type LocalStorage struct {
```
m90, line 15: Why is this "meta-type" that just embeds another type needed?

cmd/backup/storages/providers.go
```diff
@@ -0,0 +1,87 @@
...
+import (
+	"os"
+	"time"
+
+	t "github.com/offen/docker-volume-backup/cmd/backup/types"
+	"github.com/sirupsen/logrus"
+)
+
+// A pool or collection of all implemented storage provider types.
+type StoragePool struct {
```
m90, line 12: Would it be more flexible if `StoragePool` would just be a slice of things that implement a `StorageBackend` (or similar) interface and then the code just iterates over that list, calling the methods defined on the interface?

cmd/backup/storages/local.go
```diff
@@ -0,0 +1,118 @@
...
+// Specific copy function for the local storage provider.
+func (stg *LocalStorage) copy(file string) error {
+	_, name := path.Split(file)
+
+	if err := u.CopyFile(file, path.Join(stg.config.BackupArchive, name)); err != nil {
+		return fmt.Errorf("copyBackup: error copying file to local archive: %w", err)
+	}
+	stg.logger.Infof("Stored copy of backup `%s` in local archive `%s`.", file, stg.config.BackupArchive)
+
+	if stg.config.BackupLatestSymlink != "" {
```
m90, line 35: I wonder if it would be better to be more explicit about the symlinking behavior. Should this be a separate step and not be buried in `copy`? I can see pros and cons for both options so I'm not sure right now.

cmd/backup/storages/local.go
```diff
@@ -0,0 +1,118 @@
...
+	r := &LocalStorage{a}
+	return r
+}
+
+// Specific copy function for the local storage provider.
+func (stg *LocalStorage) copy(file string) error {
+	_, name := path.Split(file)
+
+	if err := u.CopyFile(file, path.Join(stg.config.BackupArchive, name)); err != nil {
+		return fmt.Errorf("copyBackup: error copying file to local archive: %w", err)
```
m90, line 31: It would be great if the error strings map back to the actual function name, i.e. this should be `copy:` or even better `(*LocalStorage).copy:`

This applies to multiple places.

m90 (commented):
Thanks for working on this, it will be very nice to have this cleaned up.

I left a couple of comments from a first read inline, and I'd also like to discuss one thing in detail here.

I think it would be great if we could change the package layout here a little bit to make it more flexible, a bit less arbitrary and also more idiomatic Go. The package layout could also guide us towards a minimal interface definition for a storage backend.

How I would envision this to look like right now would be:

```
internal/
└── storage
    ├── local
    │   └── local.go // exports a func NewLocalStorage(...params) storage.Backend
    ├── s3
    │   └── s3.go // exports a func NewS3Storage(...params) storage.Backend
    ├── ssh
    │   └── ssh.go  // exports a func NewSSHStorage(...params) storage.Backend
    ├── storage.go // exports a Backend interface type
    └── webdav
        └── webdav.go // exports a func NewWebDAVStorage(...params) storage.Backend
cmd
└── backup
    ├── archive.go
    └── .... and so on
```

Like this we'd achieve two things:

1. The script logic and control flow stays in its place without creating arbitrary boundaries and creating hard-to-answer questions about who is responsible for doing what.

2. Logic for talking to a single storage backend is encapsulated in its own package and hidden from the script.

---

I'll also check out your branch now and poke around in the code a little, trying to run examples and such. Let me know if there's anything you'd like to be clarified or discussed further.

cmd/backup/hooks.go
MaxJa4: Thank you for the detailed information.
I agree, that it makes more sense to move things like utilities in an internal package. I somehow didn't really think about that.

The only difference between the contents of `types` and `util` are, that `types` contains only structs without any logic and `util` the contrary. That was my thinking there, but it wouldn't really hurt to keep them together either.

cmd/backup/script.go
MaxJa4: Woops, I called it StorageProviders first, so that change didn't make it to the script file... should absolutely be `pool` or similar.

cmd/backup/script.go
MaxJa4: Hmm I get your point. A storage backend should have the authority to decide itself whether it shall be initialized. If it cannot do that even though it is asked to, it can still fail. While that may create duplicate code (check if should init in script and then check presence of partly the same configs again), it makes more logical sense, I agree.

cmd/backup/storages/local.go
MaxJa4: Do you mean in general or in the specific case of the `LocalStorage`?

cmd/backup/storages/providers.go
MaxJa4: I thought about that too.
If we only add backends to the pool which are initialized, we can remove the checks `if sp.S3 != nil` inside `CopyAll` and `PruneAll` since it just iterates over the available backends.

cmd/backup/storages/local.go
MaxJa4: Well, since we are already taking the code apart, I would vote for making it a separate function. Probably private and used by `copy`, but it's still a distinct enough step imo.

cmd/backup/hooks.go
m90: Another option would be not having these packages altogether. As there is only a single consumer right now, having this in different packages just creates unnecessary questions when adding new code like "where should this piece of code go" when the benefit is not really all that great.

MaxJa4:
Thank you for taking the time to review the proposed changes and describing your thoughts on them!
Like I said, I'm not too deep in Go and am a little biased from other approaches (from C#/Python), so I appreciate the input.

I totally agree, that the storages should be in their internal directory/package. You placed each backend, which is a single file, in their own subdirectory. Would you also make it a separate (internal) package? It looks odd to me to have directories for single files, but on the other hand, it cleanly separates it from `storage.go` and make it clear, what is a backend and what isn't.

I will wait with any changes in regards to your comments so we don't have conflicting changes, since you are having a look right now.

m90:
> It looks odd to me to have directories for single files

The idea is that package `storage` just defines what a `Backend` is, and then other packages (e.g. `s3` or `webdav`) implement it. It would also work if you flattened it out, but when separated it's easier not to accidentally conflate the two responsibilites.

cmd/backup/hooks.go
MaxJa4: The initial reason for me to have them both in separate packages was, that both `main` and `storages` depend on it.
If we would move `util` and `types` back into `main`, we would have cycling dependency between ´main` and `storages`.

m90: Ah ok, I didn't see that. So how about we proceed like this: after changing things as discussed, we see if we can live without an "util" type package (which would be a good sign from a design perspective I guess as it tells us things are nicely separated) and have everything in package main. If that doesn't work out, we add a single internal util package containing things that would create circular imports otherwise. Does this make sense to you too?

MaxJa4: Sounds good to me!

m90:
> I will wait with any changes in regards to your comments so we don't have conflicting changes, since you are having a look right now.

I won't be making any changes on your branch in any case so feel free to continue whenever it works for you. Don't forget to have a weekend too. Thank you!

MaxJa4:
Did some work regarding the structure like discussed:
- Internal packages
- Slice of `StorageBackend`s in `script`
- Moved conditionals for backend init back to `script`
- Stats after pruning are updated inside the respective backend itself

Utilities/util.go could not be moved back to main, as e.g. the `Join` function is used by all packages and therefore needs to stay universally accessible without causing cyclic dependencies.

.gitignore
```diff
@@ -0,0 +1,3 @@
+
+.env
```
m90, line 2: Very nitpicky: I would prefer not to have such a file that caters for a local workflow of a certain dev setup. The single artifact this repo is expected to produce is a Docker image and there is no need to ignore it. Files that are specific to a dev machine should go into its global gitignore.

internal/storage/local/local.go
```diff
@@ -0,0 +1,130 @@
...
+	u "github.com/offen/docker-volume-backup/internal/utilities"
+	"github.com/sirupsen/logrus"
+)
+
+type LocalStorage struct {
+	*strg.StorageBackend
+}
+
+// Specific init procedure for the local storage provider.
+func InitLocal(c *t.Config, l *logrus.Logger, s *t.Stats) *strg.StorageBackend {
```
m90, line 21: So the Go convention is to call such a "constructor" function `NewWhatever`.

if we also apply all of the other conventions Go has for naming things (don't stutter and so on), how about naming this `NewStorageBackend` which other packages would then call as `local.NewStorageBackend`?

internal/storage/local/local.go
```diff
@@ -0,0 +1,130 @@
...
+	t "github.com/offen/docker-volume-backup/internal/types"
+	u "github.com/offen/docker-volume-backup/internal/utilities"
+	"github.com/sirupsen/logrus"
+)
+
+type LocalStorage struct {
+	*strg.StorageBackend
+}
+
+// Specific init procedure for the local storage provider.
```
m90, line 20: Godoc comments usually start with the very name of the function, else `go vet` will complain and I think it makes sense to adhere to this. I.e. such a comment would become

```
InitLocal creates a new StorageBackend that stores archives locally
```

internal/storage/local/local.go
```diff
@@ -0,0 +1,130 @@
...
+func InitLocal(c *t.Config, l *logrus.Logger, s *t.Stats) *strg.StorageBackend {
+	a := &strg.StorageBackend{
+		Storage: &LocalStorage{},
+		Name:    "Local",
+		Logger:  l,
+		Config:  c,
```
m90, line 26: I would prefer if we wouldn't pass down the entire config object, but only the few values consumed by the specific storage implementation. It doesn't really need to know about all of the other details.

Other values like retention that apply to all backends could also be arguments for the functions.

internal/storage/local/local.go
```diff
@@ -0,0 +1,130 @@
...
+type LocalStorage struct {
+	*strg.StorageBackend
+}
+
+// Specific init procedure for the local storage provider.
+func InitLocal(c *t.Config, l *logrus.Logger, s *t.Stats) *strg.StorageBackend {
+	a := &strg.StorageBackend{
+		Storage: &LocalStorage{},
+		Name:    "Local",
+		Logger:  l,
```
m90, line 25: What do you think about just passing in a logging function instead of the entire logger object. That way, the top level command can implement logging in any way it wants to without affecting the storages.

internal/storage/s3/s3.go
```diff
@@ -0,0 +1,139 @@
...
+	"github.com/sirupsen/logrus"
+)
+
+type S3Storage struct {
+	*strg.StorageBackend
+	client *minio.Client
+}
+
+// Specific init procedure for the S3/Minio storage provider.
+func InitS3(c *t.Config, l *logrus.Logger, s *t.Stats) (*strg.StorageBackend, error) {
```
m90, line 25: I think it would be better if such a constructor function would return the interface type and not the struct type. Like this, implementation of the interface is totally up to the package.

internal/storage/storage.go
```diff
@@ -0,0 +1,46 @@
...
+
+import (
+	"time"
+
+	t "github.com/offen/docker-volume-backup/internal/types"
+	"github.com/sirupsen/logrus"
+)
+
+// Interface for defining functions which all storage providers support.
+type Storage interface {
```
m90, line 11: I'd name this type `Backend` so it reads nicely for consumers `storage.Backend`

internal/storage/s3/s3.go
```diff
@@ -0,0 +1,139 @@
...
+
+	"github.com/minio/minio-go/v7"
+	"github.com/minio/minio-go/v7/pkg/credentials"
+	strg "github.com/offen/docker-volume-backup/internal/storage"
+	t "github.com/offen/docker-volume-backup/internal/types"
+	u "github.com/offen/docker-volume-backup/internal/utilities"
+	"github.com/sirupsen/logrus"
+)
+
+type S3Storage struct {
```
m90, line 19: Those local types can be unexported, i.e. this would be `s3Storage`.

internal/storage/s3/s3.go
```diff
@@ -0,0 +1,139 @@
...
+	"errors"
+	"fmt"
+	"path"
+	"path/filepath"
+	"time"
+
+	"github.com/minio/minio-go/v7"
+	"github.com/minio/minio-go/v7/pkg/credentials"
+	strg "github.com/offen/docker-volume-backup/internal/storage"
+	t "github.com/offen/docker-volume-backup/internal/types"
```
m90, line 14: Using single letter identifiers for packages is kind of unidiomatic in Go. If there is no hard reason (e.g. collisions) to rename an import, it's probably best to use the original package name.

Dockerfile
```diff
@@ -7,6 +7,7 @@ WORKDIR /app
 COPY go.mod go.sum ./
 RUN go mod download
 COPY cmd/backup ./cmd/backup/
+COPY internal ./internal
```
m90, line 10: Can we copy these in the same step so there's still only a single Docker layer created for the source code?

cmd/backup/script.go
```diff
@@ -477,28 +392,28 @@ func (s *script) encryptArchive() error {
...
 		s.logger.Infof("Removed GPG file `%s`.", gpgFile)
 		return nil
 	})
 
 	outFile, err := os.Create(gpgFile)
-	defer outFile.Close()
 	if err != nil {
 		return fmt.Errorf("encryptBackup: error opening out file: %w", err)
 	}
+	defer outFile.Close()
```
m90, line 406: What's the reason for moving these defer statements?

m90 (commented):
Nice progress. How do you feel about the changed package layout now? I left a couple of comments inline.

In addition to that I think the next important step would be to get to the following point: Ideally, package storage just exports a single interface type and all of the constructor functions for a storage backend export a single function that returns this very interface type. In addition, the constructor functions could take non-custom types or types defined in package storage only, removing the need for the `types` package.

Let me know if have any questions. Thanks!

cmd/backup/script.go
MaxJa4: Because of an IDE warning: `should check returned error before deferring outFile.Close() (SA5001) go-staticcheck`

Dockerfile
MaxJa4: COPY only supports multiple files or one directory incl. files, not multiple directories (while also copying the directory tree structure with it).
Since it's inside the builder, the common approach would be to copy everything in there, so the complete folder structure. The `go build` only uses the go files anyway and compiles to a single binary, which is copied to the actual image. So anything we do in the builder, doesn't really matter. We could still remove some directories (instead of using everything) by using .dockerignore for that.

internal/storage/local/local.go
MaxJa4: You mean just passing a func (to the init method for example) which basically wraps the `Logger.Infof()`?

MaxJa4:
> How do you feel about the changed package layout now?

Looks good to me!

> In addition, the constructor functions could take non-custom types or types defined in package storage only, removing the need for the `types` package.

Since the storage backends don't use the Config type anymore, it can surely be moved back to main now. For the `stats` we would need to move the stats recording back to script as well, which is already the case for copy (not for prune though). So Prune would need to return its stats, so `lenCandidates` and `matches`/`len(matches)`... can you think of a better solution than that? It feels kind of messy to me. Just returning it a Stats object would need the package it lives in again.

Dockerfile
m90: True. I do like the idea of having a `.dockerignore` file that ignores everything but relevant source files. If you want to add that, it'd be cool, but we can also leave it as is.

cmd/backup/hooks.go
```diff
@@ -6,6 +6,8 @@ package main
 import (
 	"fmt"
 	"sort"
+
+	utilites "github.com/offen/docker-volume-backup/internal/utilities"
```
m90, line 10: If you use the name the package itself is using there is no need to have the mapping.

cmd/backup/script.go
m90: Ah ok. It's notoriously hard to know in which cases you need to do it before or after checking the error (sometimes you'll leak even in case of an error), so it's good the static checker figures this out for us.

m90:
> For the stats we would need to move the stats recording back to script as well, which is already the case for copy (not for prune though). 

If we already do it for copy, I think it's ok to do the same for prune, no? It's not super ugly, and if you feel it could need a bit more structure, you could also have some sort of intermediate wrapper type defined in package storage. The `stats` type defined in package `main` could then have some logic that maps the types returned from package storage without creating a circular dependency.

cmd/backup/hooks.go
MaxJa4: I know, but VS Code kept auto adding it every time I saved the file... I think it was confused about something, weird.

MaxJa4:
Did you mean with [this](https://github.com/offen/docker-volume-backup/pull/135#discussion_r928274995) and [this](https://github.com/offen/docker-volume-backup/pull/135#discussion_r928108486) comment what I implemented in c42b188 and 164a32b (last two commits)?
It guarantees a standard format and adds the formation of the context (storage backend name).

Examples:

time="2022-07-25T19:31:05+02:00" level=info msg="[S3] Uploaded a copy of backup `/tmp/backup-2022-07-25T19-31-01.tar.gz` to bucket `Testbucket-VPS`."

time="2022-07-25T19:32:36+02:00" level=error msg="Fatal error running backup: [S3] Copy: Error uploading backup to remote storage! The specified bucket does not exist: Testbucket-VPS-invalid"

m90:
> Did you mean with https://github.com/offen/docker-volume-backup/pull/135#discussion_r928274995 and https://github.com/offen/docker-volume-backup/pull/135#discussion_r928108486 comment what I implemented in https://github.com/offen/docker-volume-backup/commit/c42b188448b51517b34526d16e7dc09c7818f351 and https://github.com/offen/docker-volume-backup/commit/164a32bad6564c64437c2fb52fb13c2c4705cb17 (last two commits)?

Yes, that's pretty much what I meant, thank you.

m90:
@MaxJa4 Just to avoid any misunderstandings, can you ping me when you think this is ready to have another look at? There's no need to hurry or anything, I just wanted to make sure you're not waiting on feedback when I'm thinking you're all set. Thanks.

MaxJa4:
Sure, I will ping you!
Can I convert an issue back to a draft without loosing anything? I have never done it backwards like that, but it would help to see, that's it's still WIP (though not much left to do).

m90:
> Can I convert an issue back to a draft without loosing anything?

I doubt there is anything you can lose here, but you can also keep it a PR proper, I won't merge it by accident :) Thanks!

MaxJa4:
@m90 I think I'm done with everything I wanted to do. If there is anything left, let me know.

internal/storage/storage.go
```diff
@@ -0,0 +1,56 @@
...
+	"time"
+
+	t "github.com/offen/docker-volume-backup/internal/types"
+)
+
+// Interface for defining functions which all storage providers support.
+type Backend interface {
+	Copy(file string) error
+	Prune(deadline time.Time, pruningPrefix string) error
+}
```
m90, line 13: The fact that we still need package `types` is somehow triggering an itch of mine as it shows we don't have a "one-way flow" of responsibilities in the code. Here's a suggestion about how to maybe get rid of that:

Instead of passing in the `stats` object into the constructor functions, how about changing the interface to look like this:

```go
type Backend interface {
	Copy(file string) (CopyStats, error)
	Prune(deadline time.Time, pruningPrefix string) (PruneStats, error)
}
```

Then, the top level script could merge the returned stats on top of its own values without having to pass around a mutable object. Also we would have a clear graph of package dependencies.

What do you think?

internal/storage/storage.go
```diff
@@ -0,0 +1,56 @@
...
+
+// Generic type of storage. Everything here are common properties of all storage types.
+type StorageBackend struct {
```
m90, line 16: This is pretty academic, but I'll ask anyways: how much effort and duplication would removing this type introduce? If it's not that much I think it might be worth keeping the packages that contain the actual storage logic as isolated as possible. For example, `DoPrune` is a little odd now anyways as the most important part is passed in from the child package anyways.

internal/storage/storage.go
```diff
@@ -0,0 +1,56 @@
+package storage
+
+import (
+	"time"
+
+	t "github.com/offen/docker-volume-backup/internal/types"
+)
+
+// Interface for defining functions which all storage providers support.
```
m90, line 9: It would be great if all GoDoc comments follow the established style: https://go.dev/blog/godoc

cmd/backup/script.go
```diff
@@ -522,99 +460,19 @@ func (s *script) copyArchive() error {
...
-
-			tot, err := destination.Write(chunk[:num])
-			if err != nil {
-				return fmt.Errorf("copyBackup: error uploading the file to SSH storage: %w", err)
-			}
-
-			if tot != len(chunk[:num]) {
-				return fmt.Errorf("sshClient: failed to write stream")
-			}
+	for _, backend := range s.storagePool {
```
m90, line 470: Here lies an opportunity for adding a real world feature on top of this refactoring: as copying is a read only operation, we could now run all of the operations in parallel and considerably speed up execution times for people that back up to multiple backends. Do you think we should add this here?

In case you're interested, here's an example how it can be done: https://github.com/offen/docker-volume-backup/blob/b441cf3e2b1273462de59bcabbf0c81a6061525f/cmd/backup/exec.go#L146-L169

cmd/backup/script.go
dominikh: > It's notoriously hard to know in which cases you need to do it before or after checking the error

FWIW, Staticcheck isn't doing any "figuring out" here. It is generally agreed upon that one shouldn't return a value that needs to be closed together with a non-nil error. Any package doing so can probably be considered buggy (or at least unidiomatic). It's worth noting that in the 6 years this check has existed, nobody has complained (to me) about a false positive. On the flip side, a lot of code returns values that _mustn't_ be closed when the error is non-nil.

internal/storage/storage.go
MaxJa4: I have actually implemented exactly that already, but removed it again.
The problem I saw was, that StorageStats contains fields of Stats for each backend (S3, WebDav, ...).
Inside the `copyArchive` method of `script.go`, you would have to introduce something a switch case to check what backend is being told to copy right now so the stats get stored in the correct field. This felt like an ugly solution to me, as the goal was to remove backend-dependent code from `script.go` and now we would reintroduce some again.
Maybe reorganizing the stats for mitigating this issue would be possible, I didn't use or reverse engineer the stats and their inner workings or how they work in this project.

internal/storage/storage.go
MaxJa4: `StorageBackend` contains als fields which apply to all kinds of backends (5 fields as of now).
Not using such a generic storage backend and therefore not guaranteeing the existence of these mandatory fields, it would feel messy to me. Like it is now, you only have to add backend-specific fields to the actual backend's type struct (like AWS Bucket Name etc.), which is dependent on the backend technology.
About the DoPrune, I don't really a real opinion on that. It felt weird to do such a partial 'outsourcing' of a function to me, but if that's a common Go thing to do, I'm fine with it.

cmd/backup/script.go
MaxJa4: That's a good idea!
I would add that as a separate PR though (after this one is done ofc), as it's a standalone feature which is just being (kind of) enabled by the structural changes of this PR.

internal/storage/storage.go
m90: We could do something like this: instead of defining `StoragesStats` as a struct type, we can also use a `map[string]StorageStats`. Then, a `Name() string` method is added to `storage.Backend`. Like this, we wouldn't need a switch case or any backend dependent code and still don't need to pass down script level objects. What do you think about solving it like this?

internal/storage/storage.go
m90: > About the DoPrune, I don't really a real opinion on that. It felt weird to do such a partial 'outsourcing' of a function to me, but if that's a common Go thing to do, I'm fine with it.

It is super odd and definitely not a common Go thing to do, it was just the only way to achieve some deduplication of code in the current approach.

> Like it is now, you only have to add backend-specific fields to the actual backend's type struct (like AWS Bucket Name etc.), which is dependent on the backend technology.

The downside to this is that some hypothetical person that adds a new storage would have to learn about an additional concept on top of satisfying the one interface. But I also see your point. I'll move this around in the back of my head for a little, maybe some idea pops up.

cmd/backup/script.go
m90: Sounds good, let's do it like this,

m90:
@MaxJa4 Are you still planning on doing refinements on this or are you waiting for a review of mine? Both would be fine, I just don't want this to go stale. Thanks.

MaxJa4:
@m90 I have nothing left from my side. If we wanna optimize some small things here and there in the future, we still can with separate PRs.

internal/storage/storage.go
```diff
@@ -0,0 +1,60 @@
...
+
+import (
+	"time"
+)
+
+// Backend is an interface for defining functions which all storage providers support.
+type Backend interface {
+	Copy(file string) error
+	Prune(deadline time.Time, pruningPrefix string) (*PruneStats, error)
+	GetName() string
```
m90, line 11: `GetFoo` is not really a Go thing to do, usually you do `Foo()` and 'SetFoo(fooType)`, can we use `Name()` here instead?

internal/storage/storage.go
```diff
@@ -0,0 +1,60 @@
...
+
+// PruneStats is a wrapper struct for returning stats after pruning
+type PruneStats struct {
+	Total  uint
+	Pruned uint
+}
+
+// DoPrune holds general control flow that applies to any kind of storage.
+// Callers can pass in a thunk that performs the actual deletion of files.
+func (stg *StorageBackend) DoPrune(lenMatches, lenCandidates int, description string, doRemoveFiles func() error) error {
```
m90, line 41: For the sake of consistency I'd prefer if we used single letter receiver names throughout, i.e. this would become `s *StorageBackend`. If you're afraid of collisions with `s *script` we could also use `b` here?

This also applies to the packages themselves.

internal/storage/storage.go
```diff
@@ -0,0 +1,60 @@
...
+
+type LogType string
+
+const (
+	INFO    LogType = "INFO"
+	WARNING LogType = "WARNING"
+	ERROR   LogType = "ERROR"
+)
+
+type LogFuncDef func(logType LogType, context string, msg string, params ...interface{}) error
```
m90, line 31: `LogFuncDef` seems a little clunky, how about just `Log`?

internal/storage/webdav/webdav.go
```diff
@@ -0,0 +1,113 @@
...
+			StorageBackend: strgBackend,
+			client:         webdavClient,
+		}
+		strgBackend.Backend = webdavBackend
+		return strgBackend, nil
+	}
+}
+
+// GetName return the name of the storage backend through the interface
+func (stg *webDavStorage) GetName() string {
```
m90, line 58: Could we remove the needless indirection and define the value here once. Consumers can then just call the method instead of accessing the struct value, giving us a. a single way to access the name and b. prevent accidental mutation.

---

This applies to all packages.

.dockerignore
```diff
@@ -1 +1,8 @@
 test
+.cirecleci
```
m90, line 2: I probably could look this up myself, but as I'm here: does dockerignore support inverted statements, i.e. can we ignore everything but the source files (which is easier to keep track of)?

cmd/backup/hooks.go
m90: It'd be nice if we could still get rid of these.

m90 (commented):
I left a last round of comments, all of them very nitpicky, but probably worth it so things are kept consistent. Everything else looks pretty good now.

There's a bigger topic left in there still, but I'd like to think about this more and then possibly implement it myself: Even if it's better than passing down the top level logger, I'm not super fond of passing down that logger function, it's currently the only sideeffecfful thing being passed down and I'm not sure if we need to do this at all. Maybe we can somehow have a channel on the storage implementations that emit log messages while things are happening and then script can decide what to do about these. In any case, don't worry about this, I just wanted to ḿention that I'm still thinking about how to solve this.

cmd/backup/script.go
```diff
@@ -26,29 +29,21 @@ import (
...
-	webdavClient *gowebdav.Client
-	sshClient    *ssh.Client
-	sftpClient   *sftp.Client
-	logger       *logrus.Logger
-	sender       *router.ServiceRouter
-	template     *template.Template
-	hooks        []hook
-	hookLevel    hookLevel
+	cli         *client.Client
+	storagePool []storage.Backend
```
m90, line 41: Can we name this just `storages`` or `storageBackends` now? Pool seems a little overly complex (implying magic extra functionality) for just a list of things.

internal/storage/local/local.go
```diff
@@ -0,0 +1,137 @@
...
+func NewStorageBackend(archivePath string, latestSymlink string, logFunc storage.LogFuncDef) storage.Backend {
+
+	strgBackend := &storage.StorageBackend{
+		Backend:         &localStorage{},
+		Name:            "Local",
+		DestinationPath: archivePath,
+		Log:             logFunc,
+	}
+	localBackend := &localStorage{
+		StorageBackend: strgBackend,
```
m90, line 29: This seems a little twisted: what's the reason we need the reference back to the `strgBackend` object, creating a circular structure?

internal/storage/storage.go
```diff
@@ -0,0 +1,60 @@
...
+// Backend is an interface for defining functions which all storage providers support.
+type Backend interface {
+	Copy(file string) error
+	Prune(deadline time.Time, pruningPrefix string) (*PruneStats, error)
+	GetName() string
+}
+
+// StorageBackend is a generic type of storage. Everything here are common properties of all storage types.
+type StorageBackend struct {
+	Backend
```
m90, line 16: I don't think it's needed to embed the interface type here, is it?

internal/storage/local/local.go
m90: Initializing structs that embed other structs is a little non-obvious in Go. If I understand what you're trying to achieve here correctly, you could also do it like this:

```go
type localStorage struct {
	storage.StorageBackend
	latestSymlink string
}

// NewStorageBackend creates and initializes a new local storage backend.
func NewStorageBackend(archivePath string, latestSymlink string, logFunc storage.LogFuncDef) storage.Backend {
	return &localStorage{
		StorageBackend: storage.StorageBackend{
			Name:            "Local",
			DestinationPath: archivePath,
			Log:             logFunc,
		},
		latestSymlink: latestSymlink,
	}
}
```

Even when you declare `Name` et al on `StorageBackend` they will propagate to the top level as the type is embedded.

cmd/backup/script.go
```diff
@@ -75,7 +70,7 @@ func newScript() (*script, error) {
 		stats: &Stats{
 			StartTime: time.Now(),
 			LogOutput: logBuffer,
-			Storages:  StoragesStats{},
+			Storages:  map[string]StorageStats{},
```
m90, line 73: Not sure if this has a real world impact, but technically, this is a breaking change. Assume the script fails as early as possible and these stats have never been written to, trying to access for example `.S3` in a templatr would create a nil dereference error when previously it would just return the zero values. If we wanted to play it safe, we could define empty values for all known storage backends here, even if it creates some duplication:

```go
map[string]StorageStats{
  "S3": StorageStats{},
  // ..and so on
}
```

What do you think? Can we somehow make this smarter and easier to maintain while still keeping the old behavior?

MaxJa4:
I will have a look and implement the changes in the upcoming week. Have a nice (rest) weekend!

cmd/backup/script.go
MaxJa4: I'm not sure how to automate this, but for now it's manually initialized, just to be safe.

internal/storage/storage.go
MaxJa4: It was only needed for the DoPrune function so it is able to access the Name() func. Changed it use a function parameter now, which eliminates the need for the interface inside the struct.

internal/storage/local/local.go
MaxJa4: Thanks for the example, that indeed looks way better and simpler.

.dockerignore
MaxJa4: Unfortunately, the exception modifier doesn't support wildcards for directories like * or ** which means, that we need to manually specify the folder structure and therefore also the single backends as they have a respective folder. Explanation of that is also in the .dockerignore as comments.

MaxJa4:
Implemented all the suggested optimizations and also reimplemented #136 as it would've gotten lost due to the restructuring.
I'm done from my side.

.dockerignore
m90: It could have been so easy. Thanks for finding out.

internal/utilities/util.go
```diff
@@ -1,7 +1,7 @@
 // Copyright 2022 - Offen Authors <hioffen@posteo.de>
 // SPDX-License-Identifier: MPL-2.0
 
-package main
+package utilites
```
m90, line 4: There's a typo in the package name: `utilities`  has one more `i`.

MaxJa4: Well now I know, why VS Code insists on adding a package name to the imports :D

m90: Ah true, I didn't even think of that.

m90:
This looks really good now. As a last step, can you rebase the branch against `main` so it doesn't have conflicts?

MaxJa4:
Rebased with GitHub UI, let's see if it worked correctly. History looked good though. No conflicts anymore.

m90 (approved):
This turned out really well, thanks for all the effort put into this. I still have to make up my mind about how to get this released (so I won't merge before that), but in case you want to add other features on top of that already, let me know and we can have a `develop` branch or something similar.

Thank you very much.

m90:
There's now a `development` branch on the repository that holds this PR plus some changes of mine that I missed in review (most notably fixing the spelling of `WebDAV`).

@MaxJa4 just so that I can plan how to move ahead, are you planning to add some things on top (e.g. #138 or #101) soon? Both is fine, just so that I know.

MaxJa4:
@m90 #138 was on my list, yes, maybe #101 too. But I've been sick the last two weeks and probably will be able to start working on this by the end of next week - just so you know.

m90:
Ok, great, no need to hurry in any way. I'll let you know when this hits the main branch so you know where to start.

m90:
This is on `main` now, I really like it. Thanks again.

MaxJa4:
No worries. I like the improvements you made afterwards too, looks cleaner that way.

m90:
This is now released in [v2.22.0](https://github.com/offen/docker-volume-backup/releases/tag/v2.22.0). Thanks for your contribution.

Issue #134, opened 2022-07-21
Backup failed during copyBackup: No tomes available


MaxJa4:
* **I'm submitting a ...**
  - [ ] bug report
  - [x] feature request
  - [ ] support request

* **What is the current behavior?**

If an error occurs - specifically "no tomes available" with B2 - during copyBackup (to the remote location), there will be no retry.
The backup is deleted after the upload failed so effectively no backup was made at all.

* **If the current behavior is a bug, please provide the configuration and steps to reproduce and if possible a minimal demo of the problem.**

(Fortunately) It is rather hard to reproduce exactly that error, since it apparently means, that B2 didn't have enough available backup servers to take the backup/upload request. This only happened once for me.

* **What is the expected behavior?**

If there is a way to determine, if an error during copyBackup is one that could be mitigated by retrying, then that would be great.
At least the specific "no tomes available" could be used to trigger a retry for uploading the backup.
Also, as it looks like in the logs, the backup was deleted locally even though the upload failed. I would have expected that it at least doesn't get deleted. Maybe it could've gotten uploaded in the next iteration of the backup schedule (every night).

* **What is the motivation / use case for changing the behavior?**

Increased stability and lower chance of failed/lost backups.

* **Please tell us about your environment:**
  
  - Image version: 2.20.0
  - Docker version: 20.10.17
  - docker-compose version: 1.29.2 (docker-compose) and 2.6.0 (docker compose)

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, etc)

`Running docker-volume-backup failed with error: copyBackup: error uploading backup to remote storage: no tomes available

Log output of the failed run was:

time="2022-07-20T04:00:04+02:00" level=info msg="Created backup of `/backup` at `/tmp/backup-2022-07-20T04-00-00.tar.gz`."
time="2022-07-20T04:00:04+02:00" level=info msg="Encrypted backup using given passphrase, saving as `/tmp/backup-2022-07-20T04-00-00.tar.gz.gpg`."
time="2022-07-20T04:01:00+02:00" level=error msg="Fatal error running backup: copyBackup: error uploading backup to remote storage: no tomes available"
time="2022-07-20T04:01:00+02:00" level=info msg="Removed tar file `/tmp/backup-2022-07-20T04-00-00.tar.gz`."
[...]

PR #136, opened 2022-07-21
Print more error info for minio

MaxJa4:
Could help debugging #134 since Minio should be able to handle retrying...
The [ErrorResponse](https://pkg.go.dev/github.com/minio/minio-go/v7@v7.0.16#ErrorResponse) type contains a lot more info.

Since it's no fix and just helps finding the issue, I created the PR as draft. It may help tracking down Minio issues in the future or help the user with more verbose errors. Feel free to close this if there is no need.

### DIFF
--- a/cmd/backup/script.go
+++ b/cmd/backup/script.go
@@ -534,7 +534,7 @@ func (s *script) copyArchive() error {
 			ContentType:  "application/tar+gzip",
 			StorageClass: s.c.AwsStorageClass,
 		}); err != nil {
-			return fmt.Errorf("copyBackup: error uploading backup to remote storage: %w", err)
+			return fmt.Errorf("copyBackup: error uploading backup to remote storage: %w", minio.ToErrorResponse(err))
 		}
 		s.logger.Infof("Uploaded a copy of backup `%s` to bucket `%s`.", s.file, s.c.AwsS3BucketName)
 	}


m90:
Were you able to test this? I like the idea, but I am not sure if it actually makes a difference because of the following: if you're using the `%w` verb in the `fmt.Errorf` template, an `error` type is expected. The `error` interface is implemented on `ErrorResponse`, but it just returns the `Message` when calling `Error`, which is probably the very same result as we have right now. I.e. we'd be casting the `error` to an `ErrorResponse` and back again.

I think the proper approach here would be casting to an `ErrorResponse` and the manually constructing the error message, populating it with relevant information ourselves.

MaxJa4:
I did not experiment with it yet, since I wanted to wait for your input as you said (in the underlying issue) that you will have a look yourself too.

I can try to provoke a failed copyBackup operation. Idk yet how I would let it intentionally fail (as I would prefer to do that with a real B2 bucket to see what info is available). Perhaps setting the permissions too low so it fails, we'll see.

I will update the draft accordingly in regards to the error info printing.

m90:
> I can try to provoke a failed copyBackup operation

The easiest way I can think of would be to point the setup towards an inexistent bucket.

MaxJa4:
You were right, it had to be unpacked.
It now looks like this when using the non existant 'invalid-bucket':
`time="2022-07-22T16:00:44+02:00" level=error msg="Fatal error running backup: copyBackup: error uploading backup to remote storage: [Message]: 'The specified bucket does not exist: invalid-bucket', [Code]: NoSuchBucket, [StatusCode]: 404"`

m90 (approved):
Looks great. I'll wait with merging this until I do the next release, but it's definitely a great addition, thank you very much.

PR #161, opened 2022-10-11
add aws secret handling

pixxon:
Docker Swarm allows the user to define secrets, that are encoding and used to store secrets.
Minio already supports them, by supplying environment variables ending in _FILE.
This PR allows AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY to be configured in a similar way.

### DIFF
--- a/README.md
+++ b/README.md
@@ -196,6 +196,14 @@ You can populate below template according to your requirements and use it as you
 # AWS_ACCESS_KEY_ID="<xxx>"
 # AWS_SECRET_ACCESS_KEY="<xxx>"
 
+# It is possible to provide the keys in files, allowing to hide the sensitive data.
+# These values have a higher priority than the ones above, meaning if both are set
+# the values from the files will be used.
+# This option is most useful with Docker [secrets](https://docs.docker.com/engine/swarm/secrets/).
+
+# AWS_ACCESS_KEY_ID_FILE="/path/to/file"
+# AWS_SECRET_ACCESS_KEY_FILE="/path/to/file"
+
 # Instead of providing static credentials, you can also use IAM instance profiles
 # or similar to provide authentication. Some possible configuration options on AWS:
 # - EC2: http://169.254.169.254
@@ -945,6 +953,38 @@ services:
 
 volumes:
   data:
+```
+
+
+### Backing up to MinIO (using Docker secrets)
+
+```yml
+version: '3'
+
+services:
+  # ... define other services using the `data` volume here
+  backup:
+    image: offen/docker-volume-backup:v2
+    environment:
+      AWS_ENDPOINT: minio.example.com
+      AWS_S3_BUCKET_NAME: backup-bucket
+      AWS_ACCESS_KEY_ID_FILE: /run/secrets/minio_access_key
+      AWS_SECRET_ACCESS_KEY_FILE: /run/secrets/minio_secret_key
+    volumes:
+      - data:/backup/my-app-backup:ro
+      - /var/run/docker.sock:/var/run/docker.sock:ro
+    secrets:
+      - minio_access_key
+      - minio_secret_key
+
+volumes:
+  data:
+
+secrets:
+  minio_access_key:
+    # ... define how secret is accessed
+  minio_secret_key:
+    # ... define how secret is accessed
 ```
 
 ### Backing up to WebDAV
--- a/cmd/backup/config.go
+++ b/cmd/backup/config.go
@@ -67,7 +67,7 @@ func (c *Config) resolveSecret(envVar string, secretPath string) (string, error)
 		if err != nil {
 			return "", fmt.Errorf("resolveSecret: error reading secret path: %w", err)
 		}
-		return string (data), nil
+		return string(data), nil
 	}
 	return envVar, nil
 }
--- a/cmd/backup/config.go
+++ b/cmd/backup/config.go
@@ -4,6 +4,7 @@
 package main
 
 import (
+	"os"
 	"fmt"
 	"regexp"
 	"time"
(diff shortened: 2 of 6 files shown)


cmd/backup/config.go
```diff
@@ -58,6 +61,14 @@ type Config struct {
 	LockTimeout                time.Duration `split_words:"true" default:"60m"`
 }
 
+func (c *Config) ResolveSecret(envVar string, secretPath string) (string, error) {
```
m90, line 64: There's no need to make this method public I think, it can be called `resolveSecret` instead.

cmd/backup/config.go
```diff
@@ -58,6 +61,14 @@ type Config struct {
...
+	if secretPath != "" {
+		data, err := os.ReadFile(secretPath)
+		return string(data), err
```
m90, line 67: Most parts of the code in here will wrap returned errors with some extra info which can be pretty helpful:

```go
return string(data), fmt.Errorf("resolveSecret: error reading secret path: %w", err)
```

cmd/backup/script.go
```diff
@@ -121,10 +121,14 @@ func newScript() (*script, error) {
 	}
 
 	if s.c.AwsS3BucketName != "" {
+		AccessKeyID, err := s.c.ResolveSecret(s.c.AwsAccessKeyID, s.c.AwsAccessKeyIDFile)
```
m90, line 124: It's very uncommon to have local variable names in Go starting with an uppercase letter. Maybe `accessKeyID` and `secretAccessKey` would be more idiomatic here.

test/secret/docker-compose.yml
```diff
@@ -0,0 +1,78 @@
+# Copyright 2020-2021 - Offen Authors <hioffen@posteo.de>
```
m90, line 1: Should the directory containing the test case be called `secrets` maybe? Not sure.

m90 (commented):
This is great, thank you very much. It's not clear to me why envconfig does not want to support this (https://github.com/kelseyhightower/envconfig/issues/130), but this approach should work too.

Three things:
- I left some nitpicky comments inline, mostly about Go conventions
- Could you add this feature to the documentation as well?
- Could you try logging in to CircleCI using your GitHub credentials and see if this makes it build the PR? Right now, there is some error message I don't fully understand about how it cannot find a config file (when there is a config file clearly).

cmd/backup/script.go
```diff
@@ -121,10 +121,14 @@ func newScript() (*script, error) {
...
+		s.must(err)
```
m90, line 125: The signature here is `newScript() (*script, error)` so instead of raising a panic here it would be better to return the error instead:

```go
if err != nil {
  return nil, fmt.Errorf("newScript: error resolving secret: %w, err)
}
```

This also applies below.

test/secret/docker-compose.yml
pixxon: Not sure, I saw more singular terms in the test folder, so used the same approach.

test/secret/docker-compose.yml
```diff
@@ -0,0 +1,78 @@
...
+# SPDX-License-Identifier: Unlicense
+
+version: '3.8'
+
+services:
+  minio:
```
m90, line 7: The tests are failing right now. All that is needed to make them pass should be creating the target bucket in the `minio` service's entrypoint like in the `swarm` test:

https://github.com/offen/docker-volume-backup/blob/00c83dfac79af6f03c677e187b5bce6817b2c2a7/test/swarm/docker-compose.yml#L7-L19

(this is very non-obivous)

README.md
```diff
@@ -196,6 +196,14 @@ You can populate below template according to your requirements and use it as you
 # AWS_ACCESS_KEY_ID="<xxx>"
 # AWS_SECRET_ACCESS_KEY="<xxx>"
 
+# It is possible to provide the keys in files, allowing to hide the sensitive data.
```
m90, line 199: This is just about right, thanks :+1:

cmd/backup/config.go
```diff
@@ -58,6 +61,17 @@ type Config struct {
 	LockTimeout                time.Duration `split_words:"true" default:"60m"`
 }
 
+func (c *Config) resolveSecret(envVar string, secretPath string) (string, error) {
+	if secretPath != "" {
+		data, err := os.ReadFile(secretPath)
+		if err != nil {
+			return "", fmt.Errorf("resolveSecret: error reading secret path: %w", err)
+		}
+		return string (data), nil
```
m90, line 70:
```suggestion
		return string(data), nil
```

test/secret/docker-compose.yml
pixxon: I have made a different approach, where a sidecart is running first to create the directory. ( Troublesome, but imo its a tad bit better than entrypoint+command magic )

There was/is also some issue with minio versions, the latest one seems to store the files encoded(?), so the test couldn't just access the tar natively. I found some older version, that already supports the _FILE variables, but doesn't have this problem. Not a future proof solution, but that's all I've had time for now.

cmd/backup/config.go
pixxon: I was unsure what is preferred, since with the space my coloring showed "string" in blue ( aka type ) and without the space it showed up in orange ( aka function call )

m90: `go fmt` will remove it, so I guess that's a reasonable thing to do.

pixxon:
There is an issue with volume handling when the stack is deleted. ( Something that I ran into locally. )

When the stack is deleted, the volumes are left over and there is no flag ( as in for docker-compose ) to delete these leftover volumes. I went ahead and added a hack to make sure these are deleted, but it's not pretty.

test/secret/docker-compose.yml
```diff
@@ -0,0 +1,89 @@
...
+      - docker-volume-backup.stop-during-backup=true
+    volumes:
+      - pg_data:/var/lib/postgresql/data
+    deploy:
+      restart_policy:
+        condition: on-failure
+
+volumes:
+  backup_data:
+    name: backup_data
```
m90, line 82: If you would remove this explicit naming, would Swarm use a $PROJECT_NAME prefixed name, thus removing the need for the manual deletion of volumes you introduced?

pixxon: No, the volumes would remain. You can see it on the pg data volume, that doesn't have a special name, yet persists.

m90 (approved):
Brilliant, thank you :tophat: - I think I should get around to merging and releasing this maybe tonight or tomorrow in any case. I will let you know here when it's done.

test/secret/docker-compose.yml
pixxon: They are definietly useful for some local development, as I was running these tests locally and I ran into the volume persist issue.

pixxon:
Thanks! I will probably deploy them once they are out, the plain text passwords been annoying me for some time now.

m90:
This is now released in [v2.22.0](https://github.com/offen/docker-volume-backup/releases/tag/v2.22.0). Thanks for your contribution.

Issue #166, opened 2022-11-20
AWS Region eu-central-2 S3 buckets cannot be used to store backups


freix1:
**Describe the bug**
It appears that it is currently not possible to use the new EU (Zurich) `eu-central-2` AWS Region, that was just [released this month](https://aws.amazon.com/blogs/aws/a-new-aws-region-opens-in-switzerland), to store backups.

First, I created a S3 bucket on the new AWS Region and tried referencing it from my offen/docker-volume-backup container using the default environment variables:
```yaml
AWS_S3_BUCKET_NAME: <my_bucket_name>
AWS_ACCESS_KEY_ID: <my_aws_access_key_id>
AWS_SECRET_ACCESS_KEY: <my_aws_secret_access_key>
```
Using this configuration I received the following error when trying to create a backup:
```
time="2022-11-20T18:21:48+01:00" level=error msg="Fatal error running backup: copyArchive: error copying archive: (*s3Storage).Copy: error uploading backup to remote storage: [Message]: 'The request signature we calculated does not match the signature you provided. Check your key and signing method.', [Code]: SignatureDoesNotMatch, [StatusCode]: 403"
```
Afterwards, I added the `AWS_ENDPOINT` environment variable to see whether manually specifying the AWS Region (as specified [here](https://docs.aws.amazon.com/general/latest/gr/s3.html)) would resolve the issue:
```yaml
AWS_ENDPOINT: s3.eu-central-2.amazonaws.com
```
I then received the following error:
```
time="2022-11-20T18:35:42+01:00" level=error msg="Fatal error running backup: copyArchive: error copying archive: (*s3Storage).Copy: error uploading backup to remote storage: [Message]: 'The eu-central-2 location constraint is incompatible for the region specific endpoint this request was sent to.', [Code]: IllegalLocationConstraintException, [StatusCode]: 400"
```
Lastly, I recreated the same S3 bucket on the EU (Frankfurt) `eu-central-1` AWS Region and removed the `AWS_ENDPOINT` environment variable again. This resulted in the backups being stored correctly in the bucket without any errors.

**To Reproduce**
1. Create a new S3 bucket with the default settings on the EU (Zurich) `eu-central-2` AWS Region.
2. Reference the bucket using the default AWS S3 environment variables in a offen/docker-volume-backup container.
3. Try creating a backup with this configuration, e.g. via `docker exec <container_id> backup`.

**Expected behavior**
[...]

PR #167, opened 2022-11-21
Update MinIO Go SDK to v7.0.44

freix1:
Among other small bug fixes this update adds support for the `eu-south-2`, `eu-central-2`, `me-central-1`, `ap-southeast-3` AWS endpoints.

I have tested this update by creating a S3 bucket on the EU (Zurich) `eu-central-2` endpoint and being successfully able to store a backup to it.

Furthermore, I have also tried running the tests by replacing all occurrences of the image `offen/docker-volume-backup:${TEST_VERSION:-canary}` with my locally built image (I am not sure whether that is the correct approach). However, I got stuck on the `commands/run.sh` test case because of a reference to `/var/lib/docker/volumes/commands_archive/_data`. I cannot resolve this path on macOS and Windows (using WSL2 Ubuntu), which leads me to assume that the tests might need to be run on a Linux OS. Unfortunately, I currently do not have a Linux OS on hand, but I can set up a quick VM tomorrow and try running the tests again if need be.

This PR addresses the issue #166.


m90:
> Unfortunately, I currently do not have a Linux OS on hand, but I can set up a quick VM tomorrow and try running the tests again if need be.

CircleCI can do this for us, we'll see if the test pass very soon.

m90 (approved):
Brilliant, thanks for taking care of updating and also testing this against live infrastructure.

m90:
> Unfortunately, I currently do not have a Linux OS on hand, but I can set up a quick VM tomorrow and try running the tests again if need be.

For future reference, I usually build the image tagged `canary` myself and then run the tests, but then again I am on Linux which probably is an unfair advantage in this scenario. I don't really know how well the setup works on Mac or WSL. Pushing and having CircleCI run the tests should be a viable alternative in any case though.

m90:
This is now released in `v2.22.2`. Thanks for your contribution :tophat:

freix1:
> For future reference, I usually build the image tagged `canary` myself and then run the tests, but then again I am on Linux which probably is an unfair advantage in this scenario. I don't really know how well the setup works on Mac or WSL. Pushing and having CircleCI run the tests should be a viable alternative in any case though.

Thank you very much for the clarification. It definitely makes sense to just let CircleCI run the tests and thus avoid the trouble of having to deal with individual OS related problems.

I am very grateful for the speedy response and release as well!

PR #171, opened 2022-12-24
Add support for Azure Blob Storage

### LINKED ISSUE #101: Azure Storage Support

m90:
Closes #101


README.md
```diff
@@ -1040,6 +1058,27 @@ volumes:
...
+version: '3'
+
+services:
+  # ... define other services using the `data` volume here
+  backup:
+    image: offen/docker-volume-backup:v2
+    environment:
+      AZURE_STORAGE_CONTAINER_NAME: backup-container
+      AZURE_STORAGE_ACCOUNT_NAME: account-name
+      AZURE_STORAGE_PRIMARY_ACCOUNT_NAME: Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==
```
DevSecNinja, line 1073: Typo - should be KEY. It would be great if we could optionally specify the `AZURE_STORAGE_ENDPOINT` ourselves as many of us might use an Azure Private Endpoint or (a bit more unlikely) use something like government cloud.

m90: Good catch.

> It would be great if we could optionally specify the AZURE_STORAGE_ENDPOINT ourselves

This is already supported, but it's great you mention it because I'm not sure if it covers all the cases. Right now, the default value for `AZURE_STORAGE_ENDPOINT` is `https://{{ .AccountName }}.blob.core.windows.net/` where `.AccountName` will be replaced with the account name. When passing a custom value you could either also use the templating or just hardcode whatever you need to / want to. Would this be flexible enough to cover private endpoints and/or gov cloud scenarios?

DevSecNinja: Perfect, yes you’re spot on!

Issue #202, opened 2023-03-27
Possibility to use -u (user) option in docker exec


GeneralTao2:
**Is your feature request related to a problem? Please describe.**
I face a case where a usage of user in docker exec command option is crucial. It must be used in [Gitea dump process](https://docs.gitea.io/en-us/backup-and-restore/#using-docker-dump) to make a backup. The thing is that Gitea doesn't allow performing backup from root user, so a special user is required. I didn't manage to change user in command itself and perform a successful backup. 

**Describe the solution you'd like**
Add a `docker-volume-backup.user` label to specify user option for docker exec. If the label will be empty, then don't pass any user.

**Describe alternatives you've considered**
Nothing in mind

**Additional context**
Nothing to add

m90:
Considering the use case, I think your proposal makes sense. The only thing I wonder is whether it could be required to allow for different user values for different lifecycle values. I.e. instead of using `docker-volume.backup.user` labels, `docker-volume-backup.archive-post.user` could be used to define the user that runs `archive-post` commands. This might be a bit more tedious, but would cover all scenarios, no matter how quirky they might be.

Would you be interested in adding this feature? I'd be happy to help with reviews and and getting it released.

GeneralTao2:
Thank you. Yea, I am working on PR. It will take some time due to low Go language knowledge. I will work  on `docker-volume-backup.archive-pre.user` and `docker-volume-backup.archive-post.user` case.

m90:
Great, let me know if you need any help with that.

m90:
This is now possible as of v2.26.0: https://github.com/offen/docker-volume-backup/releases/tag/v2.26.0

PR #203, opened 2023-03-28
Possibility to use -u (user) option in docker exec

GeneralTao2:
#202 

I end up adding possibility to use user option for all the phases. I think it should make sense for process, prune and copy too.

I took the `commands/run.sh` file as a template for my test. I have:
 - removed the 7 line due to redundancy
 - took `./backup/data/dump.zip` instead of `./backup/data/dump.sql`. Gitea actually creates `dump.zip` file with configured command in `archive-pre` label
 - put sleep time to 15. in my machine it rises in around 8 sec. I put 15 just in case


GeneralTao2:
I am new in circleci. I stocked in setup menu, don't see you in organization list, or how to see the build process?

test/user/run.sh
```diff
@@ -0,0 +1,63 @@
...
+set -e
+
+cd $(dirname $0)
+. ../util.sh
+
+docker compose up -d
+sleep 15 
+
+docker compose exec backup backup
+sudo cp -r $(docker volume inspect --format='{{ .Mountpoint }}' commands_archive) ./local
```
GeneralTao2, line 12: This is questionable. What is commands_archive?

test/user/run.sh
```diff
@@ -0,0 +1,63 @@
...
+
+if [ -f ./backup/data/not-relevant.txt ]; then
+  fail "Command ran for container with other label."
+fi
+pass "Command did not run for container with other label."
+
+if [ -f ./backup/data/post.txt ]; then
+  fail "File created in post command was present in backup."
+fi
+pass "Did not find unexpected file."
```
GeneralTao2, line 28: How does this test the code? Is there any way for those files end up in ./backup/data, unless you don't put them in the folder on purpose?

m90:
> I am new in CircleCI. I stuck in setup menu, don't see you in organization list, or how to see the build process?

I recently switched this repo over to GitHub Actions, but it seems I forgot to add a trigger for pull requests. I added this to the main branch just now, so I think if you rebase your branch against `main`, CI should be triggered now. Sorry for the confusion.

m90 (commented):
The code changes look good to me already. There's two things I wanted to discuss further:

1. I'm not sure if we need to bring a full blown Gitea config just to test this feature (this creates a big burden for anyone debugging this test in the future as they still might have to find out what Gitea even is). I was imagining the test could be as simple as:
    - before doing anything, the test script creates a user fixture (sth like `testuser`) with UID 9999 in the container
    - the newly added label is used to run commands as `testuser`
    - in a `pre-archive` command, a file that contains the output of `whoami` is created, something like `whoami > whoami.txt`
    - after backing up, the test can have a look at the `whoami.txt` file and check whether it contains the expected user name
  The `commands` test case you already used as a base is a good starting point for this.
2. It'd be nice if this PR also added some documentation about the feature, i.e. extends the `Run custom commands during the backup lifecycle`  section.

Thank you very much for your work on this!

test/user/run.sh
m90: This copies the files from the docker volume named `commands_archive` to the `./local` directory. The volume is defined here https://github.com/offen/docker-volume-backup/blob/2ac1f0cea4689466ded407699519c874bba31101/test/commands/docker-compose.yml#L51 (compose adds the prefix)

test/user/run.sh
m90: The files are written by the commands under test and then archived. If the commands don't run and the files don't get created, the `-f` checks for existence above will fail.

m90:
> removed the 7 line due to redundancy

This is needed so that `pass` and `fail` print the correct test name. It will be very helpful to keep it, although it's not obvious.

m90:
Thanks for updating the branch, CI is now running every time you push to it.

test/user/run.sh
```diff
@@ -0,0 +1,31 @@
...
+set -e
+
+cd $(dirname $0)
+. ../util.sh
+current_test=$(basename $(pwd))
+
+docker compose up -d
+
+user_name=testuser
+docker exec user-alpine-1 adduser --disabled-password "$user_name"
```
GeneralTao2, line 12: seems like `docker compose exec` works slightly different, I didn't manage create a user without password, that would be accessible via exec API. I was getting `unable to find user testuser: no matching entries in passwd file` error. so I ended up creating user with `docker exec`

README.md
```diff
@@ -644,6 +644,8 @@ volumes:
 The backup procedure is guaranteed to wait for all `pre` or `post` commands to finish before proceeding.
 However there are no guarantees about the order in which they are run, which could also happen concurrently.
 
+By defould backup happens by root user. It is possible to specify a custom user in container labels with the format `docker-volume-backup.[step]-[pre|post]-[user]`. The option will allow you to run a specific step command by specified user. Make sure the user exists and present in `passwd` inside the target container.
```
m90, line 647:
```suggestion
By default the backup command is executed by the root user. It is possible to specify a custom user in container labels with the format `docker-volume-backup.[step]-[pre|post]-[user]`. The option will allow you to run a specific step command by specified user. Make sure the user exists and present in `passwd` inside the target container.
```

test/user/docker-compose.yml
```diff
@@ -0,0 +1,30 @@
...
+
+services:
+  alpine:
+    image: alpine:3.17.3
+    tty: true 
+    volumes:
+      - app_data:/tmp
+    labels:
+      - docker-volume-backup.archive-pre.user=testuser
+      - docker-volume-backup.archive-pre=/bin/sh -c 'whoami'
```
m90, line 11: Wouldn't this still need to store the output somewhere, i.e.

```
docker-volume-backup.archive-pre=/bin/sh -c 'whoami > /tmp/whoami.txt'
```

test/user/docker-compose.yml
```diff
@@ -0,0 +1,30 @@
...
+    image: offen/docker-volume-backup:${TEST_VERSION:-canary}
+    deploy:
+      restart_policy:
+        condition: on-failure
+    environment:
+      BACKUP_FILENAME: test.tar.gz
+      BACKUP_CRON_EXPRESSION: 0 0 5 31 2 ?
+      EXEC_FORWARD_OUTPUT: "true"
+    volumes:
+      - archive:/archive
```
m90, line 24: I know this is ported from the other test case, but in theory, this could also just be `./local:/archive` which would allow you to just look up the file in `./local` in the test case and you don't need to do Docker voodoo to find out about the volume location.

You can check this test case for an example:https://github.com/offen/docker-volume-backup/blob/main/test/ownership/docker-compose.yml

GeneralTao2: Done

test/user/docker-compose.yml
GeneralTao2: Done

m90 (approved):
This looks great now, thanks for the work you put into this. I'll merge and cut a new release including this tomorrow.

PR #214, opened 2023-05-08
feat: add better handler for part size

erwanlpfr:
Hello,

Here, a potential PR to allow a better handling with part size.
The problem solved is when uploading large files, the part size can be limiting.

In my case, I use S3 from scaleway and the counter size is set to 1.000 where in AWS is 10.000.
It can be problematic since Minio hard coded this. So by setting, at least, a size part I can prevent this value to reach or get over 1000.

So, what I changed :

- Adding a default value to set the part size in env vars.
- use it as parameters and compute an optimal value according to minio (which uses the counter size set to 10000).
- use it as `put` parameter to copy the file.
- also, little upgrade of packages.

I am not a go dev, so let me know if I need to change stuff.

I am testing the PR in multiple cases, so I will let you know when it works for me.
Let me know also, if you try, if it works.

EDIT : It doesn't work as it, so it is stil a draft.


erwanlpfr:
@m90 Hello, I think it works.
I was able to upload a 30 GB file into my scaleway cluster.

Let me know if you don't have any side effects !

README.md
```diff
@@ -11,47 +11,48 @@ It handles __recurring or one-off backups of Docker volumes__ to a __local direc
...
-  - [Backing up to Azure Blob Storage](#backing-up-to-azure-blob-storage)
-  - [Backing up locally](#backing-up-locally)
-  - [Backing up to AWS S3 as well as locally](#backing-up-to-aws-s3-as-well-as-locally)
-  - [Running on a custom cron schedule](#running-on-a-custom-cron-schedule)
-  - [Rotating away backups that are older than 7 days](#rotating-away-backups-that-are-older-than-7-days)
-  - [Encrypting your backups using GPG](#encrypting-your-backups-using-gpg)
-  - [Using mysqldump to prepare the backup](#using-mysqldump-to-prepare-the-backup)
-  - [Running multiple instances in the same setup](#running-multiple-instances-in-the-same-setup)
-- [Differences to `futurice/docker-volume-backup`](#differences-to-futuricedocker-volume-backup)
+- [docker-volume-backup](#docker-volume-backup)
```
m90, line 14: Are these changes to the TOC accidental?

go.mod
```diff
@@ -11,13 +11,13 @@ require (
 	github.com/gofrs/flock v0.8.1
 	github.com/kelseyhightower/envconfig v1.4.0
 	github.com/leekchan/timeutil v0.0.0-20150802142658-28917288c48d
-	github.com/minio/minio-go/v7 v7.0.44
-	github.com/otiai10/copy v1.10.0
+	github.com/minio/minio-go/v7 v7.0.52
```
m90, line 14: Is this update needed for enabling the feature?

internal/storage/s3/s3.go
```diff
@@ -17,11 +18,16 @@ import (
...
+type s3StorageInternal struct {
+	defaultPartSize int64
+}
+
 type s3Storage struct {
 	*storage.StorageBackend
 	client       *minio.Client
 	bucket       string
 	storageClass string
+	internal     s3StorageInternal
```
m90, line 30: What's the idea of introducing an `internal` type instead of just putting `defaultPartSize` on the top level `s3Storage` struct?

internal/storage/s3/s3.go
```diff
@@ -35,6 +41,8 @@ type Config struct {
 	RemotePath       string
 	BucketName       string
 	StorageClass     string
+	DefaultPartSize  int64
+	MaxUploadParts   int64
```
m90, line 45: Is this ever used?

cmd/backup/config.go
```diff
@@ -28,6 +28,7 @@ type Config struct {
 	AwsSecretAccessKey            string        `split_words:"true"`
 	AwsSecretAccessKeyFile        string        `split_words:"true"`
 	AwsIamRoleEndpoint            string        `split_words:"true"`
+	AwsDefaultPartSize            int64         `split_words:"true" default:"16"`
```
m90, line 31: Considering the currently used value on `main` is the "zero value" for `int64` I would think 0 is the correct default here, as it does exactly what is done before.

internal/storage/s3/s3.go
```diff
@@ -100,10 +111,21 @@ func (v *s3Storage) Name() string {
...
+	srcFileInfo, err := os.Stat(file)
+
+	if err != nil {
+		return fmt.Errorf("(*s3Storage).Copy: error reading the local file: %w", err)
+	}
+
+	_, partSize, _, err := minio.OptimalPartInfo(srcFileInfo.Size(), uint64(b.internal.defaultPartSize*1024*1024))
+	if err != nil {
+		return fmt.Errorf("(*s3Storage).Copy: error computing the optimal s3 part size: %w", err)
+	}
```
m90, line 123: I would think it'd be great to run all of this code only when a non-default value for `defaultPartSize` is used.

m90 (commented):
Thanks for working on this. I left a few comments, if you have any questions please inquire further. Also, CI is currently not passing when storing data in MinIO, so there seems to be something in the code that breaks the existing behavior. Let me know if you need help debugging this.

erwanlpfr:
Hello, I saw your review !
Thank you!
I will try to do so, probably next week!

Btw, about the upgrade of Minio.
It is not necessary needed but VSCode was telling me there was an update and since it is only patch,
I assumed there was no breaking change! 
Let me know if I keep that!

m90:
> It is not necessary needed but VSCode was telling me there was an update and since it is only patch,

In case it's not related to the feature, I'd prefer to stick to the version `main` is using in this PR. Like this there are as few moving parts as possible in this changeset and it's easier to figure out its consequences. MinIO can always be upgraded on its own.

erwanlpfr:
Hello !

I got time to continue,
I am actually testing cases in my side.

Let me know if everything's okay!

internal/storage/s3/s3.go
```diff
@@ -100,11 +104,26 @@ func (v *s3Storage) Name() string {
...
 	_, name := path.Split(file)
-
-	if _, err := b.client.FPutObject(context.Background(), b.bucket, filepath.Join(b.DestinationPath, name), file, minio.PutObjectOptions{
+	putObjectOptions := minio.PutObjectOptions{
 		ContentType:  "application/tar+gzip",
 		StorageClass: b.storageClass,
-	}); err != nil {
+	}
+
+	if b.partSize < 0 {
```
m90, line 112: Wouldn't this need to be

```go
if b.partSize > 0 {
```

README.md
```diff
@@ -260,6 +260,14 @@ You can populate below template according to your requirements and use it as you
 
 # AWS_STORAGE_CLASS="GLACIER"
 
+# Setting this variable will change the S3 default part size for the copy step.
+# This value is useful when you want to upload large files.
+# NB : While using Scaleway as S3 provider, be aware that the parts counter is set to 1.000.
+# While Minio uses a hard coded value to 10.000. As a workaround, try to set a higher value.
+# Defaults to "16" (MB) if unset (from minio), you can set this value according to your needs.
+
+# AWS_PART_SIZE=
```
m90, line 269:
```suggestion
# AWS_PART_SIZE=1000
```

I think it'd be good to set an example value here so people know how the number is expected to be formatted. E.g. right now I'm not sure if I'm expected to pass `1.000` or `1000`.

m90 (commented):
This looks good now, I think I spotted one logic mistake though. Let me know if that makes sense.

internal/storage/s3/s3.go
erwanlpfr: right ^^

README.md
erwanlpfr: Okay, I'll take all in consideration and I will commit as soon as I can !

erwanlpfr:
Worked ! 
So let me know if it is good for you!

m90:
This is now released as [v2.28.0](https://github.com/offen/docker-volume-backup/releases/tag/v2.28.0) - thanks for your contribution :tophat:

PR #230, opened 2023-06-17
Fix config directory typo in  README.md

jsloane:


m90:
Ouch, good catch, I hope this didn't send you down any rabbit holes. Thanks for fixing 🎩

PR #241, opened 2023-07-21
Current Docker client breaks in newer Go versions

m90:
When built with Go 1.20.6 the image starts breaking because of https://github.com/moby/moby/issues/45935


m90:
The patch is not yet released, so this branch currently still breaks the build.

PR #250, opened 2023-08-20
Fix storage backends not outputting any info logs

MaxJa4:
While implementing a new storage backend for Dropbox, I realized that any `b.Log()` **inside** the storage backends doesn't output anything. Now everything works as expected again.


m90:
Thanks for catching this :100:

PR #251, opened 2023-08-20
Add new storage backend: Dropbox (#103)

MaxJa4:
Implementation of the _official unofficial Dropbox API_ ([Repo](https://github.com/dropbox/dropbox-sdk-go-unofficial)).
Partially addresses #103.

## Implementation notes

- The Dropbox API only allows file uploads <150 MB
- Therefore, by default, a upload session is used
- Files get uploaded in 148 MB chunks (must be multiple of 4 MB)

This is fully transparent for the user and the rest of DVB. Files of up to 350GB are supported by the API this way.

## Usage notes

- It is required to create an app in Dropbox. It is also important to set the `DROPBOX_REMOTE_PATH` to the app's subfolder, where the app has access to (scope)... unless when creating the app, one chooses to give the app full access to everything.
- After adding the files modification permissions to the app scope, a new token (for `DROPBOX_TOKEN`) must be generated manually in the Dropbox app console (website).

## To be discussed

Testing this automatically is kind of an issue... for webdav/s3 etc one can simply spin up a mock service, but afaik not for Dropbox. This would require a live Dropbox account, which is kind of messy and potentially insecure.
During manual testing I verified the functionality of the file upload as well as pruning.
--> PR as draft, until this issue is addressed/discussed.

## Todo

- [x] Add example and explanation for Dropbox usage in README.md
- [x] Add mock server for testing


m90:
Thanks for working on this. As you already said I also think working against live infrastructure in tests should be avoided.

The nicest thing to have would be a mock server to work against in the tests (like it's done for Azure Blob Storage). If this does not exist, I'm not entirely sure what to do. There seems to be this: https://github.com/jgonera/fake_dropbox but it looks pretty outdated. I wonder if it would be worth searching GitHub for how others do it, as we can't be the only ones who want to test Dropbox integrations. E.g. what does `rclone` do?

MaxJa4:
`rclone` seems to use a virtual file system ([Code](https://github.com/rclone/rclone/tree/master/fstest)) for many of their storage backends, which seems quite massive and complex for this.
`fake_dropbox` seems indeed very outdated and doesn't implement the v2 API version (v1 is partially deprecated).
There is also no kind of _dummy account_ or _offline testing_ for Dropbox ([Source 1](https://www.dropboxforum.com/t5/Dropbox-API-Support-Feedback/Dummy-account-for-testing-purpose/td-p/286568), [Source 2](https://www.dropboxforum.com/t5/Dropbox-API-Support-Feedback/Offline-API-Testing/td-p/240489)).

One solution might be to write a small simple mock rest server which replies with the expected responses and errors (restricted to MkDir, DelFile, ListDir, UploadFile).

m90:
I wonder if we could use something like this https://github.com/stomakun/DropboxSwagger to automatically generate a mock server?

m90:
There also seems to be a Dropbox published spec for their API https://github.com/dropbox/dropbox-api-spec but it's using some esoteric [homegrown format](https://github.com/dropbox/stone) for defining it. Not sure how much tooling there is around it and what it would take to generate a mock server from the spec.

cmd/backup/script.go
```diff
@@ -218,6 +220,18 @@ func newScript() (*script, error) {
...
 	}
 
+	if s.c.DropboxToken != "" {
+		dropboxConfig := dropbox.Config{
+			Token:      s.c.DropboxToken,
+			RemotePath: s.c.DropboxRemotePath,
+		}
+		dropboxBackend, err := dropbox.NewStorageBackend(dropboxConfig, logFunc)
+		if err != nil {
+			return nil, err
```
m90, line 230: Nit: all other errors are wrapped (see below for an example) so we can get a poor man's stack trace.

internal/storage/dropbox/dropbox.go
```diff
@@ -0,0 +1,182 @@
...
+
+// Config allows to configure a Dropbox storage backend.
+type Config struct {
+	Token      string
+	RemotePath string
+}
+
+// NewStorageBackend creates and initializes a new Dropbox storage backend.
+func NewStorageBackend(opts Config, logFunc storage.Log) (storage.Backend, error) {
+	if opts.Token == "" {
```
m90, line 32: Considering the top level func will not even try to call this when there's no token I could live without this check personally.

internal/storage/dropbox/dropbox.go
```diff
@@ -0,0 +1,182 @@
...
+		sessionId = res.SessionId
+	}
+
+	// Send the file in 148MB chunks (Dropbox API limit is 150MB, concurrent upload requires a multiple of 4MB though)
+	// Last append can be any size <= 150MB with Close=True
+
+	const chunkSize = 148 * 1024 * 1024 // 148MB
+	var offset uint64 = 0
+
+	for {
```
m90, line 94: Is Dropbox capable of handling the upload of multiple chunks in parallel? In case yes, should we upload chunks here concurrently instead of serially?

MaxJa4:
> I wonder if we could use something like this https://github.com/stomakun/DropboxSwagger to automatically generate a mock server?

Using this with [go-swagger](https://github.com/go-swagger/go-swagger) could get us a mock server rather quickly. I will give it a try, since the `stone` format is rather arbitrairy and doesn't have a direct link to Go... only via JSON, which is already done for us by the `DropboxSwagger` repo you found.

cmd/backup/script.go
m90: I was just thinking

```go
return nil, fmt.Errorf("newScript: error creating dropbox storage backend: %w", err)
```

m90:
> Using this with [go-swagger](https://github.com/go-swagger/go-swagger) could get us a mock server rather quickly. I will give it a try, since the stone format is rather arbitrairy and doesn't have a direct link to Go... only via JSON, which is already done for us by the DropboxSwagger repo you found.

If you wanna give this a try and see if we can get a test server running, it'd be much appreciated.

internal/storage/dropbox/dropbox.go
m90: > Using one upload stream, I could saturate my ~45 MBit/s upload to Dropbox.

I would assume most times this tool is running on VPSes or similar with decent connections, so I'd rather optimize for faster connections, but then again the file size limit itself will prevent this from becoming a real bottleneck in either scenario.

MaxJa4: Seems like 140-160 MBit/s per stream is possible. In my case, 4 concurrent streams resulted in 666 MBit/s and 6 and above in 1000 MBit/s. Maybe the uplink of my server to Dropbox was the limit or they specifically limit it to 1 GBit/s... setting the default to 6 then.

MaxJa4:
Stumbled upon an auth issue: The used auth technique via token is kinda bad. The token periodically (24h?) expires, so that's no good for an automatic backup solution. OAuth2 must be used then, gotta look into that next...

m90:
Is there a way this can be done outside of this tool, e.g. by using `curl` or similar and we could just describe how it works in the README? After that people could provide the auth token to the container as before.

MaxJa4:
The access token (short-lived, only a couple hours) could be retrieved via `curl`, but it is almost the same hassle and workflow as when getting the refresh token (long-lived, no expiry) - which can be used to get infinite amounts of fresh access tokens.
I added the OAuth2 mechanism for that and created a setup manual which describes everything. Feel free to give it a quick read, if it is easy to understand.

MaxJa4:
I had a look at using the Swagger spec plus Swagger to generate a go-server that mocks Dropbox.
While doing so, I had to fix some issues with provided Dropbox Swagger spec, otherwise it wouldn't compile.
It created almost 400 files (it's a big API)... it feels like that would make this repo quite messy. I could remove all unused routes, so it's less new files, but due to the various Arg and Error types, most of them are all somewhat connected.

Seems to me like implementing this for testing is more effort than the feature itself. I will see if I can find an alternative, unless you have an idea in the meanwhile.

m90:
Another approach to use here would be to create a server that plays back a session you record from the live API once, but I'm not entirely sure how well thought out such an approach is when dealing with integration tests, instead of unit tests. 

A tool that does this for unit tests would be https://github.com/dnaeon/go-vcr for example

MaxJa4:
Stumbled upon that too. For testing purposes, the Dropbox SDK exposes the http client to be overwritten. So it should be technically be possible to implement this. I will have a further look soon.

m90:
I was actually thinking about finding something that works on integration test level (as everything else would require setting up another test mechanism in here):

- the mock server is running in a docker container and is able to either:
  - record a session against the live dropbox endpoint
  - playback a session that has previously been recorded

This way, the tests for this PR could follow the established setup.

MaxJa4:
Found something that looks promising:
https://github.com/hiredscorelabs/cornell

Will test that later.

MaxJa4:
I had a look at probably a dozen mock/replay/record server tools. Many were net suitable in regards to their deployment or features. `Cornell` was dropping headers (-> errors) and others were passing mock data which were not suitable for the supplied Swagger spec (-> errors).

The solution I came up with is now using `openapi-mock` ([Repo](https://github.com/muonsoft/openapi-mock)) which has a docker container. It can utilize provided example responses which are provided in the Swagger spec (added these manually to the `user_v2.yaml`, which I had to conver from Swagger 2.0 JSON to Swagger 3.0 YAML).

Downside of all this is, that we can't really upload anything to the mock server and therefore cannot check, if the file was correctly uploaded. But that would mostly test the API and the Dropbox library anyway, which is not our concern.

Let's see if I can get it running not only on my machine (where it worked fine) but also in the CI.

MaxJa4:
`minio Pulling 
 minio Error 
Error response from daemon: received unexpected HTTP status: 503 Service Unavailable`
Well, Docker Hub seems to be unstable today?

test/dropbox/docker-compose.yml
```diff
@@ -3,13 +3,9 @@ version: '3'
...
     image: muonsoft/openapi-mock
-    volumes:
-      - dropbox_backup_data:/var/lib/dav
-      - ./user_v2.yaml:/etc/openapi/user_v2.yaml
-      - ./openapi-mock.yaml:/etc/openapi/openapi-mock.yaml
     environment:
-      - OPENAPI_MOCK_USE_EXAMPLES: if_present
-      - OPENAPI_MOCK_SPECIFICATION_URL: https://raw.githubusercontent.com/offen/docker-volume-backup/2ea59a649be79d6cb22b8b5ad29e24377d4ab9d4/test/dropbox/user_v2.yaml
+      OPENAPI_MOCK_USE_EXAMPLES: if_present
+      OPENAPI_MOCK_SPECIFICATION_URL: https://raw.githubusercontent.com/offen/docker-volume-backup/2ea59a649be79d6cb22b8b5ad29e24377d4ab9d4/test/dropbox/user_v2.yaml
```
MaxJa4, line 8: Don't know how to bring the spec file into the CI... when it's part of the master branch, this needs to updated (points to this PR right now to get the CI working).

MaxJa4:
Finally done! Testing isn't perfect, but it will catch any errors which occur. Actual transfer cannot be tested as that would require to use Dropbox itself. That's fine for me personally.

Switching this to ready - feel free to have another look or discuss about the testing topic.

cmd/backup/script.go
```diff
@@ -156,7 +158,7 @@ func newScript() (*script, error) {
 			PartSize:         s.c.AwsPartSize,
 		}
 		if s3Backend, err := s3.NewStorageBackend(s3Config, logFunc); err != nil {
-			return nil, err
+			return nil, fmt.Errorf("newScript: error creating s3 storage backend: %w", err)
```
m90, line 161: Thanks for cleaning these up.

internal/storage/dropbox/dropbox.go
```diff
@@ -0,0 +1,267 @@
...
+
+	conf := &oauth2.Config{
+		ClientID:     opts.AppKey,
+		ClientSecret: opts.AppSecret,
+		Endpoint: oauth2.Endpoint{
+			TokenURL: tokenUrl,
+		},
+	}
+
+	isCITest := opts.Endpoint != "https://api.dropbox.com/"
```
m90, line 50: I'm a bit torn on this. Is there any way of having the code without logic that determines whether it's running tests or not? Also, what happens id someone wants to use some future Dropbx alternative with the same API, but another URL? What keeps us from mocking the token dance?

Would it be possible to call a different OAuth2 server in the tests?

internal/storage/dropbox/dropbox.go
```diff
@@ -0,0 +1,267 @@
...
+
+	logFunc(storage.LogLevelInfo, "Dropbox", "Fetching fresh access token for Dropbox storage backend.")
+	token := &oauth2.Token{RefreshToken: opts.RefreshToken}
+	if !isCITest {
+		tkSource := conf.TokenSource(context.TODO(), &oauth2.Token{RefreshToken: opts.RefreshToken})
```
m90, line 55:
```suggestion
		tkSource := conf.TokenSource(context.Background(), &oauth2.Token{RefreshToken: opts.RefreshToken})
```
There currently is no context in use and TODO would be picked up by linters. We can just use `Background` instead.

test/dropbox/openapi-mock.yaml
```diff
@@ -0,0 +1,25 @@
...
+
+# options to control generation process
+generation:
+  default_min_float: -1.073741823e+09
+  default_max_float: 1.073741823e+09
+  default_min_int: 0
+  default_max_int: 2147483647
+  null_probability: 0
+  suppress_errors: false
+  use_examples: 'if_present'
```
m90, line 25: Nit: missing new line at the end of the file.

test/dropbox/docker-compose.yml
m90: If I understand the documentation correctly you can:

a. mount the spec file into the container like `./user_v2.yml:/etc/openapi/user_v2.yml`
b. pass the path here (it's odd the param is named `_URL`, yes): `OPENAPI_MOCK_SPECIFICATION_URL: '/etc/openapi/user_v2.yaml'`

N.B.: I did not test this, it's just from the README here: https://github.com/muonsoft/openapi-mock

internal/storage/dropbox/dropbox.go
```diff
@@ -0,0 +1,267 @@
...
+
+	folderArg := files.NewCreateFolderArg(b.DestinationPath)
+	if _, err := b.client.CreateFolderV2(folderArg); err != nil {
+		switch err := err.(type) {
+		case files.CreateFolderV2APIError:
+			if err.EndpointError.Path.Tag == files.WriteErrorConflict {
+				b.Log(storage.LogLevelInfo, b.Name(), "Destination path '%s' already exists in Dropbox, no new directory required.", b.DestinationPath)
+			} else {
+				return fmt.Errorf("(*dropboxStorage).Copy: Error creating directory '%s' in Dropbox: %w", b.DestinationPath, err)
+			}
```
m90, line 108:
```suggestion
			if err.EndpointError.Path.Tag != files.WriteErrorConflict {
			    return fmt.Errorf("(*dropboxStorage).Copy: Error creating directory '%s' in Dropbox: %w", b.DestinationPath, err
			}
			b.Log(storage.LogLevelInfo, b.Name(), "Destination path '%s' already exists in Dropbox, no new directory required.", b.DestinationPath)
```

Nit: handle the error case first, and we can get rid of the `else`

internal/storage/dropbox/dropbox.go
```diff
@@ -0,0 +1,267 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("(*webDavStorage).Prune: Error looking up candidates from remote storage: %w", err)
+		}
+		entries = append(entries, res.Entries...)
+	}
+
+	var matches []*files.FileMetadata
+	var lenCandidates int
+	for _, candidate := range entries {
+		if reflect.Indirect(reflect.ValueOf(candidate)).Type() != reflect.TypeOf(files.FileMetadata{}) {
```
m90, line 237: I'm a bit lost here, but just to be sure: there is no way of doing this without having to use reflection?

internal/storage/dropbox/dropbox.go
```diff
@@ -0,0 +1,267 @@
...
+		dbxConfig.URLGenerator = func(hostType string, namespace string, route string) string {
+			return fmt.Sprintf("%s/%d/%s/%s", opts.Endpoint, 2, namespace, route)
+		}
+	} else {
+		dbxConfig.Token = token.AccessToken
+	}
+
+	client := files.New(dbxConfig)
+
+	if opts.ConcurrencyLevel < 1 {
```
m90, line 76: Good idea to check this. I would even go as far as returning an error here, so people are actually forced to provide a valid configuration instead of getting numbed by warnings.

test/dropbox/docker-compose.yml
```diff
@@ -0,0 +1,43 @@
...
+
+  backup:
+    image: offen/docker-volume-backup:${TEST_VERSION:-canary}
+    hostname: hostnametoken
+    depends_on:
+      - openapi_mock
+    restart: always
+    environment:
+      BACKUP_FILENAME_EXPAND: 'true'
+      BACKUP_FILENAME: test-$$HOSTNAME.tar.gz
```
m90, line 20: We don't really test the file being uploaded anywhere here, so I would remove these values. When removed, the `hostname` can go as well.

test/dropbox/openapi-mock.yaml
```diff
@@ -0,0 +1,25 @@
+# OpenAPI specification options
+openapi:
+  specification_url: '/etc/openapi/user_v2.yaml'
```
m90, line 3: If a configuration value for this tool is given both in the environment, and in the config file, what takes precedence? Maybe it'd be good not to define anything twice so this is easier to debug in the future.

m90 (commented):
This is great, thank you for putting so much effort into this. I left some comments inline, the only thing that I am really concerned about is the non-test-only codepath. Let me know your thoughts about it, I'm also happy to spend some more time on looking into the testing situation if you need a break.

internal/storage/dropbox/dropbox.go
m90: Alternatively, we could provide a custom `NaturalNumber` type catch such cases at config level already:

```go
type NaturalNumber int

func (n *NaturalNumber) Decode(v string) error {
	asInt, err := strconv.Atoi(v)
	if err != nil {
		return fmt.Errorf("config: error converting %s to int", v)
	}
	if asInt <= 0 {
		return fmt.Errorf("config: expected a natural number, got %d", asInt)
	}
	*n = NaturalNumber(asInt)
	return nil
}

func (n *NaturalNumber) Int() int {
	return int(*n)
}

type Config struct {
	DropboxConcurrencyLevel NaturalNumber `split_words:"true" default:"6"`
}
```

test/dropbox/user_v2.yaml
m90, line 1: I could have thought of this earlier, but does this file come with a license?

test/dropbox/docker-compose.yml
MaxJa4: I had it like that before, but I thought the CI only has the docker image and therefore no access to the repo source code.
If it has access, do you know where its scope is? Is it in the root of the project or in the respective test subfolder already? CI debugging is so time consuming unfortunately, maybe I need to install a local action runner to debug this faster.

internal/storage/dropbox/dropbox.go
MaxJa4: I didn't like it either, but it realizes any expired tokens and needs a valid one. So maybe a OAuth2 mock service can be used here yes - it won't be validated by the Dropbox mock server anyway, just the Go OAuth2 package must be happy (so it has to be able to retrieve a fresh Access Token from a given Refresh Token).

test/dropbox/docker-compose.yml
m90: In the compose file, . should resolve to the directory of the test itself.

test/dropbox/docker-compose.yml
MaxJa4: It does actually look for files with the prefix 'test' during pruning, the ListFolders command will return a (predefined) response with such a file. So it will be used. Just no modifications to the fs will be done (delete/upload).

test/dropbox/user_v2.yaml
MaxJa4: The source [repo](https://github.com/stomakun/DropboxSwagger) has `The Unlicense` as [license](https://github.com/stomakun/DropboxSwagger/blob/master/LICENSE).

internal/storage/dropbox/dropbox.go
MaxJa4: Yes, with a type switch statement. Looks less... hostile than this. Added in 'part 2' commit.

internal/storage/dropbox/dropbox.go
MaxJa4: Good idea, the configs could benefit in general from that.

MaxJa4:
Well, after 3h of trying the oauth2 mock server seems to be still allergic to Github CI... locally everything worked from the beginning like a charm, but in the CI the error 'server misbehaving' coming from the internal Docker DNS server is really weird.
Feel free the experiment yourself, I tried everything that I could come up with including Google.
Didn't find a promising alternative to that specific OAuth2 mock server either.

m90:
Thanks again. Could you maybe try to describe what the CI only problem with the OAuth Server is, so I can start investigating?

test/dropbox/user_v2.yaml
m90: Perfect. Maybe we can add a `NOTICE` file in the directory about where it's coming from and how it is licensed.

MaxJa4:
The `backup` container fails to communicate with the OAuth2 mock container.
The error message is `dial tcp: lookup oauth2_mock on 127.0.0.11:53: server misbehaving` where `127.0.0.11:53` is Docker's DNS service.
I tried fixed IPs to eliminate the DNS issue, but that didn't work. Tried the workarounds stated in the OAuth2 mock server repo located [here](https://github.com/navikt/mock-oauth2-server?tab=readme-ov-file#docker-compose).
The Swagger Mock server is referenced exactly the same way but never had issues... so I suspect that the OAuth2 mock server does something different internally.
`url.JoinPath(opts.OAuth2Endpoint, "oauth2/token")` doesn't seem to do anything bad either, as I temporarily printed the result of it out in the CI.
Locally, everything works (frustratingly).

m90:
> dial tcp: lookup oauth2_mock on 127.0.0.11:53: server misbehaving

I've recently seen this in another context where it was caused by the target service not being up and therefore unreachable, even though `depends_on` was configured. I'll see if I can make any sense of it around here.

test/dropbox/docker-compose.yml
```diff
@@ -0,0 +1,56 @@
...
+      PORT: 8090
+      JSON_CONFIG_PATH: '/etc/oauth2/config.yaml'
+    volumes:
+      - ./oauth2_config.yaml:/etc/oauth2/config.yaml
+
+  backup:
+    image: offen/docker-volume-backup:${TEST_VERSION:-canary}
+    hostname: hostnametoken
+    depends_on:
+      - openapi_mock
```
m90, line 28: I think this also needs to depend on `oauth2_mock` now?

MaxJa4: Yes it needs to. I tried that, just reverted all my (sometimes trial&error) changes.

MaxJa4:
Oh... well then I wonder how it could work on my machine. But I also saw in your commit/PR, that the file extension was wrong (I used yaml, but the contents were JSON). Maybe that led to that weird behavior.
Anyway, glad it works now! Thanks for helping here. I applied your changes to this PR.

m90 (approved):
This looks great now, thank you very much, especially for going all these extra miles on the testing topic. I'll try to release this very soon, but I'll leave it unmerged for now as for not to promise anything in the docs that isn't there yet.

m90:
This is now released in [v2.30.0](https://github.com/offen/docker-volume-backup/releases/tag/v2.30.0) :tada:

PR #261, opened 2023-08-25
Run tests Docker in Docker

### LINKED ISSUE #259: Allow running test suite Docker in Docker
The integration tests for this tool are quite extensive and very valuable, however they come with one big drawback: they are written with the assumption of a clean and otherwise unused Docker daemon in mind. If I have some random container running on my setup, tests will fail.

It would be interesting to investigate whether the test suite can be moved to run in a container that runs Docker (i.e. dind), so the tests will always have a clean slate. Ideally this would allow developers to run some script like `./test.sh` that will then run all tests in a container that spins up other containers just like it would be run in CI. Also, when creating such a `test.sh` script, filtering the tests to be run by a given pattern (e.g. `./test.sh s3` would only run the `s3` test) would be a very handy feature.

The one implementation detail I'm not sure about right now is what this means for building the image under test. Obviously, this would need to happen in the container, but I wonder if there is some way of making sure layers are cached on the host instead. Else, each test run would need to pull images and build from scratch.

m90:
Closes #259


m90:
@MaxJa4 unfortunately the backdating for the Azure test case here: https://github.com/offen/docker-volume-backup/blob/ad4e2af83fec9608ab0f610f056771f69f2f1dde/test/azure/run.sh#L49-L54

stops working when you run the test in a docker container, as you cannot change the date in a Docker container (you'd have to do it on the host). Any quick idea about how to do this differently, ideally on filesystem level? If you don't have a quick idea, no need to spend too many brain cycles on this, it's not urgent to get this merged and I can also look into it when I find the time. Thanks.

MaxJa4:
`az storage blob upload` uses the current server time as serverModified date. Which in this case is equal to the host. Setting the metadata manually via `--metadata` gets ignored, as well as the file's metadata.
I don't have an idea to trick Azure right now, costed me some time to figure out the above behavior.

m90:
Seems you crack open the Azurite server and edit the timestamp in its object database (which is just a JSON file): 99a0617a7f1639832b4d84ef071b8afdc272d478. Because what could go wrong ¯\\\_(ツ)_/¯

MaxJa4:
Seems like the most creative part of software engineering is figuring out how to test a system. Well, if it works, it works. Fine for integration testing haha. Good idea.

MaxJa4:
Saw this the other day: https://docs.docker.com/build/cache/backends/gha/
If that works with the new test setup, we maybe don't need to pull all the images anymore, only if there are new versions. At least that's how I understand it. Would reduce the 503 errors of docker hub and reduce CI run time. Just an idea :)

m90:
Yup, I am also pushing this hard at $DAY_JOB right now. Probably a good opportunity to actually try it out.

m90:
@MaxJa4 I think this is good enough to merge now, so I'll do it. In case you work on something in the future and this is causing problems our you can think of places for improvement, please let me know.

As for using the GH Actions cache, I could not figure out how to use it when pulling 3rd party images, which is the bottleneck here. Building the `docker-volume-backup` image itself is reasonably fast without a cache too.

m90:
There's also some docs here https://github.com/offen/docker-volume-backup/tree/1e39ac41f492c96fc3844cfc87ca969dded9d8a6/test

MaxJa4:
Great! Looking forward to test it out. The readme is also great to get started.

Issue #93, opened 2022-04-15
Better Pruning


Drallas:
I like this straightforward solution to backup my Docker Volumes. 

Unfortunately "The mechanism used for pruning old backups is not very sophisticated" some kind of [Policy based pruning](https://restic.readthedocs.io/en/stable/060_forget.html#removing-snapshots-according-to-a-policy) would be nice. 

Any thoughts on if it's possible / doable to implement something like this?

m90:
It's definitely possible to implement a more sophisticated mechanism, however I don't think I will work on it myself for the following reasons:
- I don't need it for my personal use of this image
- it's a tricky feature to implement (how does the API look like and how does this work against different storage backends) and making mistakes here can have very grave consequences (backups being deleted inadvertently)

That being said in case anyone else wants to pick this up and implement something like this, I am open for discussion.

---

On a side note, another possible approach would be using post-exec hooks for triggering something like rotate-backups or similar. I never tried this though, but I think it could work.

Drallas:
Thanks for tagging this as an enhancement, i would like to help, but it's a bit outside my comfort zone (zero Golang experience) to take this one on by myself.

I agree that it might be a tricky feature to implement, but perhaps the Restic [Prune](https://github.com/restic/restic/blob/1519e9f9117641872d80b0b8857920e8e16841a4/cmd/restic/cmd_prune.go) and [Forget](https://cs.github.com/restic/restic/blob/1519e9f9117641872d80b0b8857920e8e16841a4/cmd/restic/cmd_forget.go) files provide some inspiration for someone to get this started!

> Btw: My initial idea was to back up to local storage and then use Restic or Rclone scripts as some kind of (post-exec 'hook') to move the file's to B2. But it's much cleaner to have all the backup logic in the docker-compose files. Just one drawback, no pruning! 😞

simboel:
As a workaround you could execute multiple instances of this tool with different backup-filenames / prefixes. This way you could replicate some of the behaviour (like monthly backups for the last 6 years & daily backups for the last 2 months & hourly backups for the last 2 days).

This can even be done with docker-compose "overrides" (docker-compose.yml -f base.yml -f monthly.yml) where the base config is in base.yml (like GPG and S3 bucket name etc.) and monthly containing the CRON and the prefix / filename and retention configuration.

Maybe this could be added as a recipe to the README?

m90:
> As a workaround you could execute multiple instances of this tool with different backup-filenames / prefixes.

[...]

PR #262, opened 2023-08-25
Exclude specific backends from pruning

### LINKED ISSUE #207: Add option to only prune local archive
**Is your feature request related to a problem? Please describe.**

Currently it does not seem to be possible to only prune the "local" backups, which causes unwanted errors when using write-only backeds, for example S3 with the appropriate permissios.

This causes the logs to be filled with errors like:

> level=error msg="Fatal error running backup: pruneBackups: error pruning backups: [Access Denied, Access Denied, Access Denied, Access Denied, Access Denied]"

This is not good as the amount of "false positives" could make other more important errors go unnoticed due to "notification fatigue".

**Describe the solution you'd like**

I would like to be able to specify _which backends_ should be pruned, this could be achieved using a config option like `BACKUP_PRUNE_SKIP_BACKENDS=list` or by adding the option to each of the relevant backed.

Examples:
- `AWS_S3_SKIP_PRUNE` or `AWS_S3_WRITE_ONLY`
- `AZURE_STORAGE_SKIP_PRUNE` / `AZURE_STORAGE_WRITE_ONLY`
- and so on...

Ideally each backend should either implement a `ShouldPrune()` method to call [before the `Prune`](https://github.com/offen/docker-volume-backup/blob/46c6441d482094ca4be3d42c38c119f944d8f9fe/cmd/backup/script.go#L551) call, or the `Prune` itself should handle this.

**Describe alternatives you've considered**

- Run 2 instances: one performs the backups and another performs the prune (without access to the backend).
  This could work, however we need to perform the prune and the backups in very different schedules to prevent issues.
- Do not use this container for backups rotation.

**Additional context**

N/A

MaxJa4:
Fixes #207

## Content

New env variable: `BACKUP_SKIP_BACKENDS_FROM_PRUNE`
All mentioned storage backend names will be excluded from pruning.

## Notes

Tests will be added after #261 is available.

## Todo

- [x] Tests


m90:
It's fine if you wanna wait for #261, but the way tests are written won't change with that PR. It will just be easier / more robust to run them, especially locally. It could well be that it still takes me a week or two to get that into a state that I'd wanna merge it in. Anyways, whatever you prefer, just wanted to mention it.

m90:
> I saw that the following, which is in most run.sh files for testing, needs to be addressed first:

I'm surprised I didn't think of it when I set all of this up, but right now I'd implement tests for pruning like this:

- a first backup is taken
- tests on file contents are done
- the file is being backdated by ~14 days
- another backup run is triggered, configured to use ~7 retention days
- the old file should be gone, and a newer one should be there

Or am I missing something here.

> Implementing https://github.com/offen/docker-volume-backup/issues/93 with pruning policies first could fix this issue, as it should enable us to prune all existing backups (e.g. when keep-n=1).

Question is what does implementing #93 even look like? Story time: v2 of this tool is a rewrite in Go as I did not manage to properly implement pruning in the v1 bash version, instead the tool would wipe all backups (#18). Ugh. Ever since then, I tend to believe this tool is better off with a pruning strategy that is very simple and hard to get wrong.

All I am trying to say is: if we were to enhance the current mechanism, I'd like to discuss and understand it in detail. Also, I don't think the current mechanism is a real show stopper.

MaxJa4:
Sounds like a good solution to me

> the file is being backdated by ~14 days

`touch -r "file" -d '-14 days' -m "file"` would make that easy and do exactly that.

> All I am trying to say is: if we were to enhance the current mechanism, I'd like to discuss and understand it in detail. Also, I don't think the current mechanism is a real show stopper.

Since implementing #93 (in part) was mostly to fix the pruning test issue, I absolutely agree. Let's only do this when we know how and if we want to do that.

MaxJa4:
It finally works. Azure has the issue, that it ignores the modifiedTime metadata of the file and just uses the upload date instead... so going back in time, upload another 'old' file, going back to the past (set OS date) and then running backup again works.
Also added `--quiet-pull` to the first `docker  compose up` to clean up the logs a little.

Will do the other tests (where applicable) tomorrow.

MaxJa4:
All test cases for storage backends include a pruning test case now. Manipulating time was a little different for each one, but nothing magical in hindsight.

Also added two test cases in `/test/pruning` for this PR for the skip-backend-while-pruning feature. It includes a local and a S3 backend. First, the S3 backend is skipped for pruning. Then, both backends (=all available) are skipped. As all skip-functionality is in `script.go` globally and equally to all backends, IMO just using S3 as an example here is enough.

test/cli/run.sh
```diff
@@ -13,7 +13,7 @@ docker volume create app_data
 # correctly. It is not supposed to hold any data.
 docker volume create empty_data
 
-docker run -d \
+docker run -d -q \
```
m90, line 16: At least on my setup, `docker run --help` doesn't tell me that there's a `-q` option. What is it supposed to do?

test/dropbox/run.sh
```diff
@@ -6,7 +6,10 @@ cd "$(dirname "$0")"
 . ../util.sh
 current_test=$(basename $(pwd))
 
-docker compose up -d
+sudo sed -i 's/SERVER_MODIFIED_1/'"$(date "+%Y-%m-%dT%H:%M:%SZ")/g" user_v2.yaml
```
m90, line 9: This sounds as if it's a job for `envsubst`, no? Also: I think it'd be nicer if the test didn't mutate a file that is checked into source control. Instead we could output a new, gitignored file so we get a stateless test?

test/local/run.sh
```diff
@@ -52,4 +51,23 @@ if [ "$(find ./local -type f | wc -l)" != "1" ]; then
 fi
 pass "Local backups have not been deleted."
 
+# The third part of this test checks if old backups get deleted when the retention
+# is set to 7 days (which it should)
+
+BACKUP_RETENTION_DAYS="7" docker compose up -d
+sleep 5
+
+echo "## Create first backup with no prune"
```
m90, line 60: In `util.sh` there's an `info` helper for printing additional information with the context of a test. Maybe it'd be helpful to use it here?

cmd/backup/script.go
```diff
@@ -622,3 +628,14 @@ func (s *script) must(err error) {
 		panic(err)
 	}
 }
+
+// skipPrune returns true if the given backend name is contained in the
+// list of skipped backends.
+func skipPrune(name string, skippedBackends []string) bool {
```
m90, line 634: Not sure if the name should be as specific, as the logic in the function really is very generic. Maybe this could be `containedEqualFold(item string, list []string)`?

test/cli/run.sh
MaxJa4: It's short for --quiet (see [docs](https://docs.docker.com/engine/reference/commandline/run/)) and like --quiet-pull it doesn't print out the progress bar of the image layer download & extraction which spams the CI logs with 10-50 lines each time. It still prints what its doing though, especially container startup etc., only affects the progress bar of the pull.

test/dropbox/run.sh
MaxJa4: `envsubst` would need env vars to be set. The goal here not an external configuration or interaction with the environment, but to simply add the current date to the spec file, hence the `sed`.
Sure, we can do that.

test/local/run.sh
MaxJa4: Didn't realize that existed. Will add that.

cmd/backup/script.go
MaxJa4: Thought about this too, also had this briefly inside `util.go` as it is very generic and could be used by any module.

test/dropbox/run.sh
```diff
@@ -6,7 +6,11 @@ cd "$(dirname "$0")"
 . ../util.sh
 current_test=$(basename $(pwd))
 
-docker compose up -d
+cp user_v2.yaml user_v2_ready.yaml
```
m90, line 9: I'd still either:
- gitignore this file
- copy over to a temp file (you can `mktemp` to get one) and mount the tempfile instead

Else, someone will check it in at some point.

MaxJa4: Oh, completely forgot that. Working on two branches simultaneously and multiple things didn't work well this time. Will add it.

m90 (approved):
Brilliant work once again, thanks!

PR #263, opened 2023-08-25
Cleanup: Lint warnings and deprecated packages

MaxJa4:
- `ioutil.ReadFile` is officially deprecated and is now called `os.ReadFile`
- `ioutil.ReadAll` is officially deprecated and is now called `io.ReadAll`
- Other minor lint warning fixes -> repo is lint (go-staticcheck) warning free 🎉 
- Replaced deprecated `golang.org/x/crypto/openpgp` (with security and stability issues) with suggested, maintained fork `github.com/ProtonMail/go-crypto/openpgp`
- ran `go mod tidy` which switched to oauth2 package from indirect to direct (bc of Dropbox I would assume)
- Added `golangci-lint` as separate workflow with the linter `staticcheck` (VS Code standard) enabled


cmd/backup/exec.go
```diff
@@ -51,19 +51,15 @@ func (s *script) exec(containerRef string, command string, user string) ([]byte,
 		outputDone <- err
 	}()
 
-	select {
```
MaxJa4, line 54: Not quite sure what the purpose of the select is here.
Because there is only one case, it will block/wait anyway -> remove the select and just reading straight from the channel.
This also means, that the go func right over this doesn't do anything as we wait here anyway.
Or do I miss something?

m90: If I recall correctly, most of this code is copied verbatim from the test suite of `moby` (see the license header). I don't know too much about it, and I'd maybe just leave it unchanged?

Your reasonsing sounds correct though.

m90:
> Other minor lint warning fixes -> repo is lint (go-staticcheck) warning free 

Do you think it would make sense to run `go-staticcheck` in CI?

> Replaced deprecated golang.org/x/crypto/openpgp (with security and stability issues) with suggested, maintained fork github.com/ProtonMail/go-crypto/openpgp

Reading [the issue](https://github.com/golang/go/issues/44226) I am a bit lost about what package is actually supposed to do. Who is suggesting to use the `protonmail` maintained one?

MaxJa4:
> Do you think it would make sense to run `go-staticcheck` in CI?

Yes that would make sense. Usually linting is a part of CI or at least some linting bot in GH. 


> Reading [the issue](https://github.com/golang/go/issues/44226) I am a bit lost about what one is actually supposed to do. Who is suggesting to use the `protonmail` maintained one? 

This is what I was referring to (last paragraph of the linked issue):

> Forks abound. Every major player that actually relies on OpenPGP and Go maintains a fork (https://github.com/keybase/go-crypto, https://github.com/ProtonMail/go-crypto, ...). Developers would be better served by finding those, and by centralizing efforts onto one, without blocking on the resource limits of the Go team

Protonmail is the most stared and most active fork, also used in their production setup.

m90:
> Yes that would make sense. Usually linting is a part of CI or at least some linting bot in GH.

If you'd wanna add this in this PR, I'd appreciate. I guess I would put it in a dedicated workflow instead of prepending it to the test run.

MaxJa4:
Added [golangci-lint](https://github.com/marketplace/actions/run-golangci-lint) as separate workflow.
Selected linters are currently only `staticcheck` as this is used by default in VS Code.
I'm done here from my side.

.golangci.yml
```diff
@@ -0,0 +1,5 @@
+linters:
+  # Enable specific linter
+  # https://golangci-lint.run/usage/linters/#enabled-by-default
+  enable:
+    - staticcheck
```
m90, line 5: I think now that we have this, we could also enable `govet` here?

.github/workflows/golangci-lint.yml
```diff
@@ -0,0 +1,54 @@
...
+          # skip-cache: true
+
+          # Optional: if set to true, then the action won't cache or restore ~/go/pkg.
+          # skip-pkg-cache: true
+
+          # Optional: if set to true, then the action won't cache or restore ~/.cache/go-build.
+          # skip-build-cache: true
+
+          # Optional: The mode to install golangci-lint. It can be 'binary' or 'goinstall'.
+          # install-mode: "goinstall"
```
m90, line 54: Missing new line at end of file.

.golangci.yml
m90, line 5: Missing new line at end of file.

.github/workflows/golangci-lint.yml
```diff
@@ -0,0 +1,54 @@
...
+          # working-directory: somedir
+
+          # Optional: golangci-lint command line arguments.
+          #
+          # Note: By default, the `.golangci.yml` file should be at the root of the repository.
+          # The location of the configuration file can be changed by using `--config=`
+          # args: --timeout=30m --config=/my/path/.golangci.yml --issues-exit-code=0 
+
+          # Optional: show only new issues if it's a pull request. The default value is `false`.
+          only-new-issues: true
```
m90, line 41: Considering the repo is currently lint free, this could also be `false`. Not sure if it really makes a difference?

.github/workflows/golangci-lint.yml
```diff
@@ -0,0 +1,54 @@
+name: Run Quality Tests
```
m90, line 1:
```suggestion
name: Run Linters
```
Super pedantic: using `Quality` sounds to me as if linting ensures quality software. Which is I think is a bit more complicated than that...

.golangci.yml
MaxJa4: Seems like by default there are also `errcheck, gosimple, ineffassign, unused` enabled. I kept them on.

MaxJa4:
Dev hint: with this you can pre-check if the linters (will) pass in the CI: ([Source](https://golangci-lint.run/usage/install/#docker))
```docker
docker run -t --rm -v $(pwd):/app -w /app golangci/golangci-lint:v1.54.2 golangci-lint run -v
```

The linter CI mentions the location of any issues in the CI logs (`Overview` -> `Annotations` - [Example](https://github.com/offen/docker-volume-backup/actions/runs/5991710042)).

m90:
Is it somehow possible to make VSCode (the Go extension rather) pick up the configuration from `golanglint-ci.yaml`?

m90:
Looks like it: https://golangci-lint.run/usage/integrations/

m90 (approved):
Thanks :tophat:

PR #266, opened 2023-09-02
Replace Gzip with PGzip

### LINKED ISSUE #265: Use more efficient drop-in replacements for compression
## Is your feature request related to a problem?

Currently, `compress/gzip` is the default compression option for compatibility reasons, since the way faster `zstd` is not compatible with that file format.
From the same author from whom we use the `zstd` library (klauspost) exist more compression libraries, which are optimized versions of their standard implementation.

I tested these two drop-in replacements for `compress/gzip`:
- `github.com/klauspost/compress/gzip` -> [optimized gzip](https://pkg.go.dev/github.com/klauspost/compress/gzip) implementation, still as specified in gzip standard [spec](https://www.rfc-editor.org/rfc/rfc1952.html)
- `github.com/klauspost/pgzip` -> [parallelized version](https://github.com/klauspost/pgzip) of the optimized gzip, therefore also compatible

<details><summary>Test setup</summary>
<p>

**HW**: i7 13700KF, 32GB RAM, PCIe 4.0 NVMe SSD with ~7 GB/s read (sequential)
**SW**: Win10, Go v1.21.0, klauspost pkgs v1.16.7
**Parameters**: Compression level = 4 for all (default for zstd)
**Test data used**: Full Silesia Corpus contents ([link](https://sun.aei.polsl.pl/~sdeor/index.php?page=silesia))
**Methodology**: Read file (tar'ed test data), de-/compress in memory (no file writing)

</p>
</details> 

<details><summary>Results</summary>
<p>

| Package        | File size after | Compression | Decompression |
|------------|-----------|-------------|---------------|
| Gzip orignal | 33,54%    | 2152ms      | 863.1ms       |
| Gzip klauspost  | 34.29%    | 1249ms      | 711.4ms       |
| PGzip klauspost | 34.37%    | 246.5ms     | 674.0ms       |
| Zstd for reference | 31.89%    | 672.0ms     | 198.6ms       |

</p>
</details> 

## Describe the solution you'd like

To make existing `.tar.gz` / `gzip` setups benefit from vastly superior compression speeds (~1.7x with Gzip, ~8.7x with PGzip) and slightly improved decompression speeds while preserving compatibility, I'm suggesting using one of the optimized `gzip` implementations as drop-in replacement.
Note, that it might not be desirable in an automated/unsupervised backup scenario to use all available CPU power for compression like with `PGzip`'s parallelism (level can be adjusted though), as live production workloads could be influenced by a backup run. The optimized `Gzip` is still single threaded afaik.
While we're at it, maybe it makes sense to introduce three compression presets: fast, default (level 4 like here and default for all packages afaik), best. The user can then prioritize downtime (fast preset) or storage size/cost (best preset). Not set would be default like now.

If you see the need for more testing, feel free to suggest more.

MaxJa4:
Closes #265 

Replaces `compress/gzip` with `github.com/klauspost/pgzip` (drop-in replacement, compatible) to significantly increase compression (archiving) performance (1.5-8+ times depending on setting `GZIP_PARALLELISM`).
Default is `1` to not use up all available compute capacity on production servers during the archiving process.

---

@m90: Also added the `--timeout 3` for `docker compose down` commands in the test suite.


MaxJa4:
New test fails with `/bin/sh: /code/test/pgzip/run.sh: Permission denied` right at the beginning. It works locally. Do you know what is happening here @m90?

m90:
Sorry for not mentioning this before, the `run.sh` file needs to be executable, i.e. you need to `chmod +x run.sh` still.

cmd/backup/archive.go
```diff
@@ -63,7 +64,11 @@ func compress(paths []string, outFilePath, subPath string, algo string) error {
 	prefix := path.Dir(outFilePath)
 	switch algo {
 	case "gz":
-		compressWriter = gzip.NewWriter(file)
+		w := pgzip.NewWriter(file)
+		if err := w.SetConcurrency(1<<16, concurrency); err != nil {
```
m90, line 68: If I understand https://github.com/klauspost/pgzip/blob/17e8dac29df8ce00febbd08ee5d8ee922024a003/gzip.go#L21-L25 correctly, the default block size is `1<<20`. Why did you choose a different value? I maybe missing something, but this https://github.com/offen/docker-volume-backup/issues/265#issuecomment-1703898701 also sounds as if the default should be used.

cmd/backup/config.go
```diff
@@ -148,3 +149,21 @@ func (n *NaturalNumber) Decode(v string) error {
 func (n *NaturalNumber) Int() int {
 	return int(*n)
 }
+
+type NaturalNumberZero int
```
m90, line 153:
```suggestion
type WholeNumber int
```
Maybe that's a bit less clunky?

MaxJa4:
> the run.sh file needs to be executable
It was executable by default locally, but I had to run `git add --chmod=+x run.sh` to make git aware of that. Additionally GitHub Desktop sees a change after that, but when committing it says "error: no changes". Doing everything with command line and VS Code worked though.

cmd/backup/archive.go
MaxJa4: My bad, don't know how I came up with 16 there. It's supposed to be 20. Fixed that
Also changed the compression level to 5 like mentioned in the issue #265 since Gzip default (4) seems to correspond to PGzip level 5 in terms of compression ratio. So backups don't get bigger suddenly.

m90:
> but I had to run git add --chmod=+x run.sh to make git aware of that

Strange, usually git should pick this up all by itself. Could it be that you set `core.filemode` to `false` somehow?

Anyways, good to see it's running as expected now.

cmd/backup/config.go
```diff
@@ -148,3 +149,21 @@ func (n *NaturalNumber) Decode(v string) error {
...
+
+type WholeNumber int
+
+func (n *WholeNumber) Decode(v string) error {
+	asInt, err := strconv.Atoi(v)
+	if err != nil {
+		return fmt.Errorf("config: error converting %s to int", v)
+	}
+	if asInt < 0 {
+		return fmt.Errorf("config: expected a natural number, got %d", asInt)
```
m90, line 161:
```suggestion
		return fmt.Errorf("config: expected a whole number, got %d", asInt)
```

cmd/backup/script.go
```diff
@@ -503,7 +504,12 @@ func (s *script) createArchive() error {
 		return fmt.Errorf("createArchive: error walking filesystem tree: %w", err)
 	}
 
-	if err := createArchive(filesEligibleForBackup, backupSources, tarFile, s.c.BackupCompression.String()); err != nil {
+	concurrency := s.c.GzipParallelism.Int()
+	if concurrency == 0 {
+		concurrency = runtime.GOMAXPROCS(0)
```
m90, line 509: Can this have possible side effects on following operations? Also, should it only apply when `gz` is the target compression format? Right now, this would also apply when using `zstd` compression.

I.e. should this happen further down the call stack in `compress` itself?

MaxJa4: Preferably, I'd like to put the `switch` case in `archive.go` for selecting and initializing the respective compression backend in its own function to unclutter the compress function there. It should only care about compression, not what backend is used there, as everything after the initialization is the same. Just to use the opportunity to clean up in there.

MaxJa4:
I renamed `WholeNumber` again as its logically false. The name suggests that any whole number (including negative ones) are allowed - which is not the case. Since its exactly like `NaturalNumber` but with a zero, I just named it `NaturalNumber0` now to align it a little with the mathematical convention. I also added a doc string for both types to avoid confusion.

m90:
Hmm, I got all my knowledge from [Wikipedia](https://en.wikipedia.org/wiki/Natural_number) which says whole numbers are natural numbers + 0. Then again, it's probably disputed and not really well defined.

MaxJa4:
> Could it be that you set core.filemode to false somehow?

Not that I know of. I just changed to true, but now GIT says I have changes in all .sh files... so it definitely picked something up and was turned off before.

> which says whole numbers are natural numbers + 0

Seems like I found exactly the opposite side of that discussion haha. Well, I don't have a real preference here, now that the doc string is there to explain it.

test/util.sh
```diff
@@ -35,6 +35,9 @@ docker() {
         up)
           shift
           command docker compose up --timeout 3 "$@";;
+        down)
+          shift
+          command docker compose down --timeout 3 "$@";;
```
m90, line 40: I'm curious how this works. If a timeout for the containers is set on `up`, why does `down` need to know of it as well?

m90 (approved):
:bullettrain_side:

Issue #270, opened 2023-09-10
Unable to run entrypoint.sh when specifing a rootless user with docker arguments


kitsumed:
**Describe the bug**
Cannot specific a rootless docker user to run the repo due to entrypoint.sh access (root) requirement.
Error : `/bin/sh: can't open '/root/entrypoint.sh': Permission denied`
Seem to be due to how [dockerfile](https://github.com/offen/docker-volume-backup/blob/336c5bed71c1505e16ac05ac89134ace7354ed21/Dockerfile) generate the image.

**To Reproduce**
1. Create a docker container / stack
2. Set the docker user argument to a user & group without root permissions

**Expected behavior**
entrypoint.sh should be run by docker

**Version (please complete the following information):**
 - Image Version: v2.33.0
 - Docker Version: 24.0.5

**Additional context**
Not running the app with root is limiting in what you can do, however in my case, I does not need to run it as root, as such, I should be able to use a rootless user & group.

m90:
The user inside the container has to be root as otherwise it couldn't interact with the Docker socket being mounted from the outside. This is a key feature of this image so I think it's not really worth finding some smart configuration that allows for both as then people will try running the rootless version but expect it to be able to interact with Docker.

Also see https://github.com/offen/docker-volume-backup/issues/36

If you really need this, I think the easiest thing to do is create an image off this one where you shuffle around files and run with a different user. Be aware that you cannot use any Docker facing feature in this case.

kitsumed:
In my current use case, I does not need to bind to a docker socket, making the use of a root user kind of useless. But I does understand your point.

Else you could detect if the entrypoint is running rootless with and show a warning telling them docker socket won'Ttwork.

m90:
If the entrypoint script was moved to `/usr/bin` or similar, would that suffice to let you pass a different `--user` to Docker?

kitsumed:
> If the entrypoint script was moved to `/usr/bin` or similar, would that suffice to let you pass a different `--user` to Docker?

Might be worth trying it, I can't right now trought.

m90:
@kitsumed Not sure yet how/if this will be merged, but here's a PR that demonstrates what changes would need to be done in the Dockerfile to enable your use case #272 - maybe it helps you with deriving a custom Docker image from the existing one.

kitsumed:
[...]

PR #272, opened 2023-09-16
Allow running backup as non-root user in container

m90:
This is a proof of concept for #270

Not sure yet whether I really want to merge this.


m90:
Closing this as it's not really addressing all of the problems with running as non-root.

PR #288, opened 2023-10-20
docs: Update the custom commands documentation

### LINKED ISSUE #287: Documentation issues with custom commands
**Describe the bug**

Documentation issues with custom commands:-

1. The docker socket must be mounted for custom commands to work (needs to see labels) - the socket is not just for stopping containers.
2. The docker-compose labels are seen across the whole docker estate, so unless you use unique `EXEC_LABEL`s, should you have multiple `docker-volume-backup` sets, you will end up running *all* the custom commands across the estate even those within entirely different docker-compose configurations

nigelm:
Include requirement for docker socket and the need to use EXEC_LABEL if you are running multiple copies

Fixes #287


nigelm:
This attempts to deal with the issues I encountered when using custom commands

m90 (approved):
LGTM, thank you very much!

PR #319, opened 2023-12-20
Add method for converting object to JSON in template

hywax:
This adds the `toJSON` template function to notifications, allows the template data to JSON serialized and adds a built in default JSON template (simply `{{ . | toJSON }}`).

Useful if want to format a message in a third-party system according to certain rules.

### DIFF
--- a/cmd/backup/notifications.go
+++ b/cmd/backup/notifications.go
@@ -6,6 +6,7 @@ package main
 import (
 	"bytes"
 	_ "embed"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"os"
@@ -84,7 +85,8 @@ var templateHelpers = template.FuncMap{
 	"formatBytesBin": func(bytes uint64) string {
 		return formatBytes(bytes, false)
 	},
-	"env": os.Getenv,
+	"env":    os.Getenv,
+	"toJSON": toJSON,
 }
 
 // formatBytes converts an amount of bytes in a human-readable representation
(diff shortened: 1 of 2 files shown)


cmd/backup/notifications.go
```diff
@@ -106,3 +108,12 @@ func formatBytes(b uint64, decimal bool) string {
 	}
 	return fmt.Sprintf(format, float64(b)/float64(div), "kMGTPE"[exp])
 }
+
+func toJSON(v interface{}) string {
```
m90, line 112: Looking at what [package sprig](https://masterminds.github.io/sprig/) does, I wonder if we should have `toJSON` (plain) and `toPrettyJSON` (indented) so that users that require a single line for whatever reason have their needs catered as well.

m90 (commented):
This is a great addition, thank you. I left a comment about a possible improvement inline, lmk what you think.

cmd/backup/notifications.go
hywax: If the user wants JSON, they don't care if it's one line or not. Since the user will work with the data after the JSON parser.

m90: I'm mostly thinking about the usecase of outputting NDJSON and then feeding this elswhere.

E.g. you could have a template like this

```
{"Type": "local", "Data": {{ .Stats.Storages.Local | toJSON }} }
{"Type": "s3", "Data": {{ .Stats.Storages.S3 | toJSON }} }
```
and get one line per storage. If the JSON output is always pretty printed, this use case wouldn't work as you'd get more than one line per item.

hywax: I can only see 2 uses for it:
1) Template only - `{{ . | toJSON }}`
2) JSON to JSON - `{"Type": "s3", "Data": "{{ .Stats.Storages.S3 | toJSON }}" }`

That said, I think trying to create JSON based on a template is a very rare case study.

But perhaps you as the creator of the library know better where to develop the product. But then again, I don't think it's necessary here.

m90: I think it'd make sense to provide both pretty printed as well as standard json methods, so if you could still add this to your changes, it'd be much appreciated.

hywax: Made 2 separate functions: `toJSON` and `toPrettyJson`.

m90:
This is now released in [v2.35.0](https://github.com/offen/docker-volume-backup/releases/tag/v2.35.0) thank you :tophat:

PR #323, opened 2024-01-07
Fix dropbox doc link

peteward44:
Just a minor PR to fix a little irritation I found that the link in the dropbox documentation is broken, and the dropbox setup md file has a typo so i fixed that too! 👍


m90:
Renaming the file theoretically breaks external links to this very page as the UR changes, but I guess it's better to fix the typo now, so let's do it. Thanks for spotting and fixing.

PR #328, opened 2024-01-13
Stop swarm services by scaling down and up

### LINKED ISSUE #327: Improve Docker Swarm support
Docker Swarm is supported by this tool, albeit there's a caveat: services need to use a restart policy of `on-failure` where `always` is not supported. This brings problem like single node clusters deleting services on reboot.

It would be nice if we could find a more "swarmy" way of temporarily stopping a service. Maybe it could be done by scaling down the service to 0 replicas or even removing and redeploying the service entirely.

I'm creating this ticket mostly to collect my own thoughts around this feature (which is something I would like to use myself).

m90:
Closes #327

### DIFF
--- a/cmd/backup/script.go
+++ b/cmd/backup/script.go
@@ -30,10 +30,10 @@ import (
 	"github.com/ProtonMail/go-crypto/openpgp"
 	"github.com/containrrr/shoutrrr"
 	"github.com/containrrr/shoutrrr/pkg/router"
+	"github.com/docker/cli/cli/command/service/progress"
 	"github.com/docker/docker/api/types"
 	ctr "github.com/docker/docker/api/types/container"
 	"github.com/docker/docker/api/types/filters"
-	"github.com/docker/docker/api/types/swarm"
 	"github.com/docker/docker/client"
 	"github.com/leekchan/timeutil"
 	"github.com/offen/envconfig"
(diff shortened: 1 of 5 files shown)


cmd/backup/script.go
```diff
@@ -318,14 +318,107 @@ func newScript() (*script, error) {
...
 	if s.cli == nil {
 		return noop, nil
 	}
 
+	dockerInfo, err := s.cli.Info(context.Background())
+	if err != nil {
+		return noop, fmt.Errorf("stopContainers: error getting docker info: %w", err)
+	}
+	isDockerSwarm := dockerInfo.Swarm.LocalNodeState != "inactive"
+	if isDockerSwarm {
```
pixxon, line 339: I would recommend making it stop both the services and the containers.
It might be a bit of an odd approach, but it is possible to create containers in a docker swarm that are not bound to services.

pixxon:
Can I offer any kind of help to get this pr resolved faster?

cmd/backup/script.go
m90: > I would recommend making it stop both the services and the containers.

I'm not entirely sure I understand. If I stop containers in Swarm Mode, Swarm will try to restart them, unless the service has a `on-failure` restart policy. This is what's currently happening when you use this image with Swarm. From what I understand you cannot really "stop" a service, you can a. scale it down to 0 or b. delete and redeploy.

This is currently not reflected in the code, but I was thinking the best (also least breaking) approach here would be to do the following:

- if the label is on the container (i.e. not in `deploy`), the container will be stopped and restarted
- if the label is put on the service (i.e. in `deploy`), the service will be scaled down and up again

If you want to help out here, that's more than welcome, although I would still like to understand the pros and cons of all approaches (stopping containers, removing services, scaling services) better so we can pick the best one.

pixxon: Sorry, I was not clear with the exact issue by not stopping the containers too.

Even if a node is part of a Docker Swarm, while unorthodox, it is still possible to create standalone containers on it by calling `docker run`.
Given the following scenario:
1. Node is part of a Docker Swarm
2. User for some reason calls `docker run` on the specific node
3. Applies `docker-volume-backup.stop-during-backup` labels on the *container*
4. Starts the backup process
5. At this point, the backup will skip stopping containers, since it believes the containers are all part of a service, which is a false assumption.

To avoid this happening, I would recommend stopping both services *and* containers. The only error case here would be if a container has the label and the service that it belongs to also has it, so a container would be stopped twice. ( Which is a configuration error, but can be ignored by the backup if needed. )


Regarding your understanding of the services, that is correct. You cannot stop a service. but scale it down to 0.

The main difference between stopping all containers and scaling down is where the containers will be scheduled after the backup. In the first case they will stay on the same node, there is a guarantee that the container will not move to a new one. For scaling, it will depend on how the service is configured, if the placement requirements allow, it is possible to schedule it on another node. This might have some issues if the services configuration is faulty. ( And the volumes are local. )

The advantage of the option to configure on service labels, at least for me, is reducing the chances of messing up the configuration. For containers, one has to apply the `on-failure` restart policy and put the labels on the containers, not the services. The latter is a bit counterintuitive, since they are usually on the services, eg Traefik.

Removing and recreating the services sounds a bit too intrusive for my taste and I am not totally sure if you can save the state of the service/stack for redeployment. As a user of this product, I would prefer the scaling alternative instead.


I like the idea you have explained, it is also something that I would implement. It would not break the previous configurations upon upgrade, but allows the tool to be configured in a more Swarm friendly way. Also having both options would allow the user to pick based on their exact scenario.

m90: Ah, now I get it, also makes sense seeing you commented on this particular line. I think that should give us a pretty straightforward plan of what's to add (scale services to zero and up again when they are labeled). I'll see if I can try this out the next couple of days and will keep you posted here. Thanks for your input.

m90:
I'm continuing work on this in #328 which currently is still very much a work in progress. I'll let you know how it goes.

Issue #160, opened 2022-10-10
Container that provides network service to other containers should be restarted before others


reggaemahn:
**Describe the bug**
Currently, it seems like the sequence of stopping and starting containers during backup is somewhat arbitrary. This causes an issue if one or more of the containers are using another container as the network provider and the dependent container is started before the network provider container. In this case the dependent containers fail to start. 

**To Reproduce**
1. Create 3 containers - A, B & C in a docker-compose file
2. Set `network` to `service:A` for container B & C
3. Run the backup

**Expected behavior**
Container A is stopped, backed up and restarted. Then containers B & C should be stopped and started again.

**Desktop (please complete the following information):**
 - Image Version: v2.21.0
 - Docker Version: 20.10.17
 - Docker Compose Version (if applicable): 2.4

**Additional Context**
The _Steps to reproduce_ above may or may not work depending on whether container A is down when either B or C is being restarted. I'm not quite sure how to have it fail reliably.

m90:
I didn't even know you can use a container as a "network provider" and I also cannot find any info about this in here: https://docs.docker.com/compose/networking/

How does this compare to using an explicitly declared network that you put your containers in? Like this, stopping and restarting should work as you expect it. Would this be a viable workaround for your situation?

---

As for the root cause of what you describe it seems as if Docker will give no guarantees about the ordering of the containers returned here: https://github.com/offen/docker-volume-backup/blob/00c83dfac79af6f03c677e187b5bce6817b2c2a7/cmd/backup/script.go#L268-L274

which causes the behavior you describe. I have a hard time imagining some mechanism that sorts these correctly based on their dependencies (this is pretty complex and might involve more than just networks) without blowing up complexity a whole lot. If this was to be supported I guess the way to go is having using label their services with some sort of priority value that is then used to sort the container before starting and stopping.

reggaemahn:
Unfortunately for my use case a separate network won't work. 

You can see the reference for this network mode [here][1] and [here][2]. One example of using this when you have a container that connects to a VPN and other containers connected to it have all their traffic routed through that VPN connection. [Example][3] Wireguard image that can be used for something like this. 

You're right though. I had a quick look at the docs and there doesn't seem to be an easy way to get the dependencies with the docker cli, mostly because `depends_on` etc. is compose syntax. `docker-compose` does do it internally ([reference][4]). 

[...]

PR #333, opened 2024-01-25
Improve Swarm support

### LINKED ISSUE #327: Improve Docker Swarm support

m90:
Supersedes #328 

Closes #327


m90:
@pixxon So I think this should work as discussed in #328 now. I still have to do cleanup and add documentation but if you have feedback or even want to test an image built off this branch yourself, I'd be thankful for any remarks.

The [test](https://github.com/offen/docker-volume-backup/blob/swarm-scale/test/services/docker-compose.yml) is currently the only reference for how to configure things, but I think you have an idea of how it's supposed to work anyways.

cmd/backup/script.go
```diff
@@ -318,44 +319,79 @@ func newScript() (*script, error) {
...
+	filterMatchLabel := fmt.Sprintf(
 		"docker-volume-backup.stop-during-backup=%s",
 		s.c.BackupStopContainerLabel,
 	)
+
+	allContainers, err := s.cli.ContainerList(context.Background(), types.ContainerListOptions{})
+	if err != nil {
+		return noop, fmt.Errorf("(*script).stopContainersAndServices: error querying for containers: %w", err)
+	}
 	containersToStop, err := s.cli.ContainerList(context.Background(), types.ContainerListOptions{
```
pixxon, line 354: What happens if a user labels a container with `docker-volume-backup.stop-during-backup` which is part of a service and also adds the same label to the service itself?

cmd/backup/script.go
```diff
@@ -421,21 +500,55 @@ func (s *script) stopContainers() (func() error, error) {
 			}
 		}
 
-		if len(restartErrors) != 0 {
+		var scaleUpErrors []error
+		if isDockerSwarm {
+			for _, service := range servicesToScaleDown {
```
pixxon, line 505: Shouldn't this be `scaledDownServices`? If a the backup tried to scale a service down, but it ran into an error, what will be the state of it? If the PreviousSpec is not updated, we might make unintentional change to Replica count

cmd/backup/script.go
```diff
@@ -421,21 +500,55 @@ func (s *script) stopContainers() (func() error, error) {
...
+
+				if updatedService.PreviousSpec == nil {
+					scaleUpErrors = append(
+						scaleUpErrors,
+						errors.New("(*script).stopContainersAndServices: service does not have PreviousSpec, cannot scale back up."),
+					)
+					continue
+				}
+
+				updatedService.Spec.Mode.Replicated.Replicas = updatedService.PreviousSpec.Mode.Replicated.Replicas
```
pixxon, line 520: as mentioned above, not sure if backup can rely on PreviousSpec. Would also be a possibility to store the Replica count in a variable and update to that. ( Or maybe a deep copy of the original spec and that can be compared to before trying to scale up. If spec what is expected does not match previousspec, there was somebody doing something in the background?

pixxon:
Except the three comments above, the feature looks great to me. I will build an image and later today try it out. 

I have one scenario in mind, where the service does not scale down in time. It might leave the service in a state that requires manually intervention from the user.

cmd/backup/script.go
m90: I was wondering the same, yes. Documentation says do one or the other, so one option would be to say "tough luck, please do what is described". Another option would be checking if a container has a parent service, and in case it does check if it is labeled as well. In case there is a conflict, bail.

Probably worth doing this extra check.

m90: What I think would happen in practice is:
- the container is stopped
- its parent service is scaled down, removing the container
- the script tries to restart a now-removed container and fails

cmd/backup/script.go
m90: > If a the backup tried to scale a service down, but it ran into an error, what will be the state of it? 

The idea was to always scale back up services, even if scaling down might have failed (this is also happening for containers), in order to try the very best to restore the previous state. If scaling down failed and we try to update a service to its current state, nothing bad will happen anyways.

You're right about how relying on PreviousSpec only is maybe a bit too risky. I'll think about how to make this a bit more robust.

cmd/backup/script.go
m90: >  there was somebody doing something in the background?

Yeah, that's a valid concern. I'd probably do it like you describe and wrap the service object like

```go
type handledService {
    serviceID                                       string
    replicaCountBeforeScaleDown uint64
}
```

Like this, the script can lookup the current spec at scale up time and only change the replica count to what it was before scaling down. This effectively allows users to change everything _but_ the replica count while a backup is performed. I have a hard time thinking of an approach that allows updating the desired replica count while a backup is happening. But I guess this is ok, as long as it's mentioned in the docs.

m90:
Thanks for your review. I'll address these points and some other tiny bits I was thinking of later today, and I'll also push a pre-release to Docker Hub, so I can test it myself in a setup of mine. I'll let you know when this is out.

cmd/backup/script.go
pixxon: If possible I would aim to be user friendly, while it would be hard to accidentally configure it this way, still a possibility.

Instead of checking if the container has a parent service with the labels, the other way might be easier to implement. ( Check if a labelled service has any also labelled containers. )

cmd/backup/script.go
pixxon: That a good reasoning and then consistent with containers. I am not sure what happens if scale down fails and then we try to scale up. Either way creates an error that should be looked at by the user.

m90:
I pushed a build of `ba095ea` tagged as `v2.36.0-alpha.1` (both to Docker Hub and GHCR)  and also already tested it in a low-stakes Swarm setup of mine, where it seems to work as intended. I'll let it run unsupervised for a while now before proceeding.

If anyone wants to test this in their setup as well, I'm happy about any feedback. Documentation is not yet deployed, but can be accessed here https://github.com/offen/docker-volume-backup/blob/swarm-scale/docs/how-tos/use-with-docker-swarm.md

cmd/backup/script.go
```diff
@@ -369,38 +436,92 @@ func (s *script) stopContainers() (func() error, error) {
...
+				continue
+			}
+
+			for _, warning := range response.Warnings {
+				s.logger.Warn(
+					fmt.Sprintf("The Docker API returned a warning when scaling down service %s: %s", service.Spec.Name, warning),
+				)
+			}
+
+			if err := progress.ServiceProgress(context.Background(), s.cli, service.ID, discardWriter); err != nil {
```
m90, line 476: It occurs to me this returns slightly earlier than containers have actually been removed, creating potential races between volume consumers (in my setup the app still holds the lock on a SQLite database, making the backup fail).

If I put a short sleep of a second after this line, it works, but obviously this shouldn't be the proper solution. Maybe this should instead check if there are no more containers that are children of this service or similar.

I'll need to look into this still.

m90: It seems adding

```go
			// progress.ServiceProgress returns too early, so we need to manually check
			// whether all containers belonging to the service have actually been removed
			for {
				containers, _ := s.cli.ContainerList(context.Background(), types.ContainerListOptions{
					Filters: filters.NewArgs(filters.KeyValuePair{
						Key:   "label",
						Value: fmt.Sprintf("com.docker.swarm.service.id=%s", service.ID),
					}),
				})
				if len(containers) == 0 {
					break
				}
				time.Sleep(time.Second)
			}
```
instead of sleeping actually works, although I am not sure if this is expected or if I am just using `ServiceProgress` incorrectly.

m90:
I just pushed `v2.36.0-alpha.2` which seems to fix the issue and can be tested by others.

cmd/backup/script.go
pixxon: The related command `docker service scale` will converge a bit before the actual containers are removed. Seems like the related API calls behave the same way.

cmd/backup/script.go
```diff
@@ -369,38 +436,111 @@ func (s *script) stopContainers() (func() error, error) {
...
-	if len(stopErrors) != 0 {
-		stopError = fmt.Errorf(
-			"stopContainers: %d error(s) stopping containers: %w",
-			len(stopErrors),
-			errors.Join(stopErrors...),
-		)
+	var scaledDownServices []swarm.Service
+	var scaleDownErrors []error
+	if isDockerSwarm {
+		for _, svc := range servicesToScaleDown {
```
m90, line 442: I'm wondering if both scaling down and scaling up multiple services should happen concurrently as it can take quite a while for a single service to converge, making downtimes during backup a bit longer than needed.

Then again we already have problems like described in https://github.com/offen/docker-volume-backup/issues/160

Any opinions on this @pixxon ?

pixxon: Since after scale down there is a check to ensure that every container is fully removed, it should be possible to initiate the scale down on every service in a detached mode. ( I am not sure if the go api has this feature, but the cli does. )

Regarding the scale up, it could also be done in a detached mode, since backup can move on from there. Only thing that could be helpful is to check if every service is up to expected number before a timeout limit. And if not, issue a warning so the user is aware.

m90: As for __scaling up__: I'm not sure if detaching would actually make this faster. If you look at the code in Docker CLI: https://github.com/docker/cli/blob/abf8cff233f6d3bfb93d0c15f5a55f1987418ae3/cli/command/service/scale.go#L80-L88 it seems detaching skips what we are doing here when calling `progress.ServiceProgress`. However, the loop that polls for no containers to be present anymore will apparently always takes longer than `progress.ServiceProgress`. So all that "detaching" (i.e. skipping `progress.ServiceProgress`) would give is us fewer meaningful errors (as returned by `progress.ServiceProgress` and the possibility of getting stuck in an endless loop).

As for __scaling down__: Detaching would indeed make things considerably faster, then again this would not allow notification features to kick in in the case of failures to scale back up. Maybe a simple solution would be to run the entire "restart" function in a goroutine so that encryption, upload et al can happen in parallel.

pixxon: I think the detach would still help you with the scaling up, since the order of the scales and waits would be different. ( In attached mode: `scale wait scale wait scale wait` in detached mode: `scale scale scale wait wait wait`.

To me really the only difference is that if one of the services fail to scale down, we could terminate earlier. But the users should not rely on this, since it is non-deterministic.

m90: > scale scale scale wait wait wait

This is exactly what we'd get if every operation on a service would be running in its own goroutine (as implemented in cfa5f073d2b31b3b64e7d81108f4ce905885119a)

pixxon: Then I am not exactly sure why you mean it would not make it faster. The reordering should allow Docker to work in the background while we move on to the next service.

Regarding the endless loop, I think one reasonable solution would be to set a timeout limit, after which the backup will just assume there is an error. Sadly I don't think we can really do much about this, since Docker itself will just have a container stuck in Terminating or some other state for a longer period of time.

m90: > Then I am not exactly sure why you mean it would not make it faster. 

I think this is mostly a misunderstanding. It's running in parallel now in any case, not trying to bail in case a service fails, but trying to finish all of them. I think that's a bit easier to understand.

> Regarding the endless loop, I think one reasonable solution would be to set a timeout limit, after which the backup will just assume there is an error. 

Added it like this with a 5 minute timeout.

m90:
316a7a6ab188a5c7b122b64bd187347bbaa514db is now pushed as `v2.36.0-alpha.3`

pixxon:
Small note about the configuration. Users can apply the label via `BACKUP_STOP_CONTAINER_LABEL` environment variable. However in case of __services__ it might be a bit misleading.

I can't really think of a great solution to avoid this. Adding `BACKUP_STOP_SERVICE_LABEL` would allow the user to set separate labels, but would increase the complexity of the configuration, allowing more user errors to happen. ( User sets wrong label and environment variable combination so the services do not stop... )

m90:
> Users can apply the label via BACKUP_STOP_CONTAINER_LABEL environment variable. However in case of services it might be a bit misleading.

I was already thinking about the same problem, and since you also raise it, it's probably a good idea to do something about it.

I think the proper way would be to deprecate `BACKUP_STOP_SERVICE_LABEL` (i.e. using it logs a warning and it shall be removed in any potential future major version) and start using `BACKUP_STOP_DURING_BACKUP_LABEL`, which is also more in line with the labels themselves.

m90:
I pushed 87ea8d0930dd6ee36f24abc7e21971bb1804a3ad as `v2.36.0-alpha.4` and will deploy it to all of my setups next. In case this works as expected, I think this is ready to go now.

m90:
This is now released as [v2.36.0](https://github.com/offen/docker-volume-backup/releases/tag/v2.36.0). Thanks for the input @pixxon

PR #338, opened 2024-01-31
Move cron scheduling inside application

### LINKED ISSUE #268: Consider using a dedicated cron package
The limitations of BusyBox's cron implementation when it comes to defining schedules crop up again and again e.g. #267 #185 #117

There seems to be a standalone apk package for cron https://pkgs.alpinelinux.org/package/edge/main/x86/apk-cron which seems lightweight enough to install it in the image if:
- it supports all syntax that is supported by BusyBox
- it adds missing features like `\` or `-`

Since this __the core feature__ of this image, possible incompatibilities should be scrutinized closely before actually implementing.

pixxon:
Closes #268

Allows the application to take control of the scheduling instead of relying on crond.
- Tries to read config from environment variables
- Reads all the files under `/etc/dockervolumebackup/conf.d` and each creates a new job
- Without `--foreground` flag, returns to old behaviour

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,4 +18,4 @@ RUN apk add --no-cache ca-certificates
 COPY --from=builder /app/cmd/backup/backup /usr/bin/backup
 COPY --chmod=755 ./entrypoint.sh /root/
 
-ENTRYPOINT ["/root/entrypoint.sh"]
+ENTRYPOINT ["/usr/bin/backup", "--foreground"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,4 +18,4 @@ RUN apk add --no-cache ca-certificates
 COPY --from=builder /app/cmd/backup/backup /usr/bin/backup
 COPY --chmod=755 ./entrypoint.sh /root/
 
-ENTRYPOINT ["/usr/bin/backup", "--foreground"]
+ENTRYPOINT ["/usr/bin/backup", "-foreground"]
--- a/cmd/backup/configProvider.go
+++ b/cmd/backup/configProvider.go
@@ -1,0 +1,80 @@
+// Copyright 2021-2022 - Offen Authors <hioffen@posteo.de>
+// SPDX-License-Identifier: MPL-2.0
+
+package main
+
+import (
+	"fmt"
+	"log"
+	"os"
+
+	"github.com/joho/godotenv"
+	"github.com/offen/envconfig"
+)
+
+type envProxy func(string) (string, bool)
+
+func loadConfig(lookup envProxy) (*Config, error) {
+	envconfig.Lookup = func(key string) (string, bool) {
+		value, okValue := lookup(key)
+		location, okFile := lookup(key + "_FILE")
+
+		switch {
+		case okValue && !okFile: // only value
+			return value, true
+		case !okValue && okFile: // only file
+			contents, err := os.ReadFile(location)
+			if err != nil {
+				log.Panicf("newScript: failed to read %s! Error: %s", location, err)
+				return "", false
+			}
+			return string(contents), true
+		case okValue && okFile: // both
+			log.Panicf("newScript: both %s and %s are set!", key, key+"_FILE")
+			return "", false
+		default: // neither, ignore
+			return "", false
+		}
+	}
+
+	var c = &Config{}
+	if err := envconfig.Process("", c); err != nil {
+		return nil, fmt.Errorf("newScript: failed to process configuration values: %w", err)
+	}
+
+	return c, nil
+}
+
+func loadEnvVars() (*Config, error) {
+	return loadConfig(os.LookupEnv)
+}
+
+func loadEnvFiles(folder string) ([]*Config, error) {
+	items, err := os.ReadDir(folder)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to read files from env folder: %w", err)
+	}
+
+	var cs = make([]*Config, 0)
+	for _, item := range items {
+		if item.IsDir() {
+			log.Println("Skipping directory")
+		} else {
+			envFile, err := godotenv.Read(".env")
+			if err != nil {
+				log.Println("Error reading file: %w", err)
+			}
+			lookup := func(key string) (string, bool) {
+				val, ok := envFile[key]
+				return val, ok
+			}
+			c, err := loadConfig(lookup)
+			if err != nil {
+				log.Println("Error loading config from file: %w", err)
+			}
+			cs = append(cs, c)
+		}
+	}
+
+	return cs, nil
+}
--- a/cmd/backup/configProvider.go
+++ b/cmd/backup/configProvider.go
@@ -5,8 +5,9 @@ package main
 
 import (
 	"fmt"
-	"log"
+	"log/slog"
 	"os"
+	"path/filepath"
 
 	"github.com/joho/godotenv"
 	"github.com/offen/envconfig"
@@ -25,12 +26,12 @@ func loadConfig(lookup envProxy) (*Config, error) {
 		case !okValue && okFile: // only file
 			contents, err := os.ReadFile(location)
 			if err != nil {
-				log.Panicf("newScript: failed to read %s! Error: %s", location, err)
+				slog.Error(fmt.Sprintf("failed to read %s", location), "error", err)
 				return "", false
 			}
 			return string(contents), true
 		case okValue && okFile: // both
-			log.Panicf("newScript: both %s and %s are set!", key, key+"_FILE")
+			slog.Error(fmt.Sprintf("both %s and %s are set!", key, key+"_FILE"))
 			return "", false
 		default: // neither, ignore
 			return "", false
@@ -39,30 +40,37 @@ func loadConfig(lookup envProxy) (*Config, error) {
 
 	var c = &Config{}
 	if err := envconfig.Process("", c); err != nil {
-		return nil, fmt.Errorf("newScript: failed to process configuration values: %w", err)
+		slog.Error("failed to process configuration values", "error", err)
+		return nil, err
 	}
 
 	return c, nil
 }
 
 func loadEnvVars() (*Config, error) {
+	slog.Info("loading config from environment variables")
 	return loadConfig(os.LookupEnv)
 }
 
 func loadEnvFiles(folder string) ([]*Config, error) {
+	slog.Info(fmt.Sprintf("loading config from environment files from directory %s", folder))
+
 	items, err := os.ReadDir(folder)
 	if err != nil {
-		return nil, fmt.Errorf("Failed to read files from env folder: %w", err)
+		slog.Error("failed to read files from env folder", "error", err)
+		return nil, err
 	}
 
 	var cs = make([]*Config, 0)
 	for _, item := range items {
 		if item.IsDir() {
-			log.Println("Skipping directory")
+			slog.Info("skipping subdirectory")
 		} else {
-			envFile, err := godotenv.Read(".env")
+			p := filepath.Join(folder, item.Name())
+			envFile, err := godotenv.Read(p)
 			if err != nil {
-				log.Println("Error reading file: %w", err)
+				slog.Error(fmt.Sprintf("skipping subdirectory %s", p), "error", err)
+				return nil, err
 			}
 			lookup := func(key string) (string, bool) {
 				val, ok := envFile[key]
@@ -70,7 +78,8 @@ func loadEnvFiles(folder string) ([]*Config, error) {
 			}
 			c, err := loadConfig(lookup)
 			if err != nil {
-				log.Println("Error loading config from file: %w", err)
+				slog.Error(fmt.Sprintf("error loading config from file %s", p), "error", err)
+				return nil, err
 			}
 			cs = append(cs, c)
 		}
--- a/cmd/backup/main.go
+++ b/cmd/backup/main.go
@@ -24,7 +24,10 @@ func runBackup(c *Config) {
 	defer func() {
 		s.must(unlock())
 	}()
-	s.must(err)
+	if err != nil {
+		log.Println("Error during locking:", err)
+		return
+	}
 
 	defer func() {
 		if pArg := recover(); pArg != nil {
@@ -37,7 +40,7 @@ func runBackup(c *Config) {
 						fmt.Sprintf("An error occurred calling the registered hooks: %s", hookErr),
 					)
 				}
-				os.Exit(1)
+				return
 			}
 			panic(pArg)
 		}
@@ -49,7 +52,7 @@ func runBackup(c *Config) {
 					err,
 				),
 			)
-			os.Exit(1)
+			return
 		}
 		s.logger.Info("Finished running backup tasks.")
 	}()
@@ -94,7 +97,7 @@ func main() {
 				log.Println("Added cron job with schedule: ", c.BackupCronExpression)
 				_, err := cr.AddFunc(c.BackupCronExpression, func() { runBackup(c) })
 				if err != nil {
-					log.Println("Failed to create cron job with schedule: ", c.BackupCronExpression)
+					log.Println("Failed to create cron job with schedule:", c.BackupCronExpression)
 				}
 			}
 		} else {
@@ -102,7 +105,7 @@ func main() {
 				log.Println("Added cron job with schedule: ", c.BackupCronExpression)
 				_, err := cr.AddFunc(c.BackupCronExpression, func() { runBackup(c) })
 				if err != nil {
-					log.Println("Failed to create cron job with schedule: ", c.BackupCronExpression)
+					log.Println("Failed to create cron job with schedule:", c.BackupCronExpression)
 				}
 			}
 		}
--- a/cmd/backup/main.go
+++ b/cmd/backup/main.go
@@ -4,12 +4,18 @@
 package main
 
 import (
+	"flag"
 	"fmt"
+	"log"
 	"os"
+	"os/signal"
+	"syscall"
+
+	"github.com/robfig/cron/v3"
 )
 
-func main() {
-	s, err := newScript()
+func runBackup(c *Config) {
+	s, err := newScript(c)
 	if err != nil {
 		panic(err)
 	}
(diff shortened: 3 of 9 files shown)


pixxon:
Still heavily in progress, but wanted to push the current state.

Since I am not too familiar with golang, all sort of feedback is more than welcome.

cmd/backup/main.go
```diff
@@ -4,19 +4,28 @@
...
 	"os"
+	"os/signal"
+	"syscall"
+
+	"github.com/robfig/cron/v3"
 )
 
-func main() {
-	s, err := newScript()
+func runBackup(c *Config) {
```
m90, line 17: I guess this function will now have to return an error, so that we can:
- exit 1 when not running in the foreground
- log/notify, but not exit when running in the foreground

cmd/backup/main.go
```diff
@@ -4,19 +4,28 @@
...
+	s, err := newScript(c)
 	if err != nil {
 		panic(err)
 	}
 
 	unlock, err := s.lock("/var/lock/dockervolumebackup.lock")
 	defer s.must(unlock())
-	s.must(err)
+	if err != nil {
+		log.Println("Error during locking:", err)
```
m90, line 26: This raises an interesting question: so far a backup run has an exclusive lock mostly because I was too lazy to deal with the consequences of concurrent invocations. Do you think we should keep it like this or remove this limitation?

m90 (commented):
Thanks for starting to work on this. I'll leave comments here and there, but please let me know when you think this is ready for a proper review.

cmd/backup/main.go
pixxon: I would keep it for now, I do not see a solution to create a more fine grade solution. Worst case is that a job gets blocked by another for some time.

It could be moved into seperate issue, but I am not totally sure how locks could be more finer graded. A container could use multiple volumes each with their own schedule. It would be possible that both want to stop/start at the same time...

cmd/backup/main.go
pixxon: I tried to wrap my head around how `defer` and all that works in Go. I think there was a bug in the code regarding the lock, we instantly unlocked it in the next statement. 
> [The deferred call's arguments are evaluated immediately, but the function call is not executed until the surrounding function returns.](https://go.dev/tour/flowcontrol/12)

I will go and do some restructure for the rest of the codebase by removing any reference to `panic` ( and the related `must` function. )

Do you have any rule on how to split what is sent into the script logger ( `s.logger` ) and what is fine using the global `log` object? ( I believe the only different is that the first one will provide some stats regarding individual script runs? )

m90: > I tried to wrap my head around how defer and all that works in Go. I think there was a bug in the code regarding the lock, we instantly unlocked it in the next statement.

Very good catch, seems this was broken here https://github.com/offen/docker-volume-backup/pull/263/files and I did not notice in review. I'll patch this directly as it's kind of bad.

> Do you have any rule on how to split what is sent into the script logger

The current setup is mostly a remnant of a time when the standard library did not have structured logging and package `logrus` was used (changed in https://github.com/offen/docker-volume-backup/pull/247).

Currently, the idea is to always use the logger instance (because this is how logrus works), but I am not sure if this is worth it. The nice thing about doing it like this is obviously testability, then this repo doesn't have any unit tests and only integration tests (it would be nice to have unit tests though, but I simply never find the time to add them). In any case I think we should pick one of the two approaches and stick to it always, i.e. always use the logger instance or always use the global one. Is there anything that makes it especially hard to always use the logger instance in your changeset?

m90: Fixing the locking in https://github.com/offen/docker-volume-backup/pull/339

pixxon: > Is there anything that makes it especially hard to always use the logger instance in your changeset?

Currently the logger instance is bound to the `script` object, which is used for one execution. This is also used to collect and provide logs to the notifications. This feature demands the use of an individual logger instance for every backup run. 

My idea to stay consistent through the whole application:
- On startup create the logger that will be used through the lifetime of the application.
- Inject this into every backup run, so they can wrap around it and provide both logs to stdout and into the buffer.

I think this would result in a consistent experience ( every log message is done via `logger.Info` ) and keep the same functionality.

m90: Sounds good to me :+1:

cmd/backup/main.go
```diff
@@ -4,19 +4,32 @@
...
 	"os"
+	"os/signal"
+	"syscall"
+
+	"github.com/robfig/cron/v3"
 )
 
-func main() {
-	s, err := newScript()
+func runBackup(c *Config) *error {
```
m90, line 17: `error' is an interface type

```go
// The error built-in interface type is the conventional interface for
// representing an error condition, with the nil value representing no error.
type error interface {
	Error() string
}
```

so it's rather uncommon to return a pointer, else you might end up with a pointer to a pointer.

I.e. this can be `func runBackup(c *Config) error`

pixxon: I already changed it, the reason I had to was due to the defered unlock call which itself could return an error.

pixxon: I decided to make #341 where the logging could be standardized and improved in the whole application. Right now the `main` loop will use the logger initialized at startup.

cmd/backup/main.go
pixxon: I decided to make #341  where the logging could be standardized and improved in the whole application. Right now the main loop will use the logger initialized at startup.

cmd/backup/main.go
```diff
@@ -65,4 +82,79 @@ func main() {
...
+					logger.Error("unexpected error during backup", "error", err)
+				}
+			})
+			if err != nil {
+				logger.Error("failed to create cron job", "schedule", c.BackupCronExpression)
+			}
+		}
+
+		cs, err := loadEnvFiles(*envFolder)
+		if err != nil {
```
m90, line 115: This will need to be more specific. There are two types of errors `loadEnvFiles` can return:
- the directory simply does not exist (which is expected)
- something failed (e.g. permissions error or malformed content) when trying to read the files, which should make the script stop

The most straight forward thing to do would be to stat the directory in `loadEnvFiles` and return [`os.ErrNotExist`](https://pkg.go.dev/os#pkg-variables) in case it does not exist. `func main` can then check which error is returned and proceed accordingly.

cmd/backup/main.go
```diff
@@ -4,21 +4,33 @@
...
 	unlock, err := s.lock("/var/lock/dockervolumebackup.lock")
+	if err != nil {
+		return err
+	}
+
 	defer func() {
-		s.must(unlock())
+		err = unlock()
+		if err != nil {
+			ret = err
```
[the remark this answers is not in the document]
pixxon: The defered function will be executed after the last statement in the function, but before it goes up the call stack. Due to this, using the named return variable, it is possible to modify the return inside a defered function. I hopefully read the right docs about this, [Deferred functions may read and assign to the returning function’s named return values
](https://go.dev/blog/defer-panic-and-recover)
Also tried it out with an example in [playground](https://go.dev/play/p/_ELXY1m437O).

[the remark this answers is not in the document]
m90: Wow, that's crazy, didn't know about this yet.

m90:
Do you think it would make sense to already update the `entrypoint.sh` script so that tests also run the non-cron scheduler?

pixxon:
> Do you think it would make sense to already update the `entrypoint.sh` script so that tests also run the non-cron scheduler?

Yesterday I ran the whole test suite with success locally using the new scheduler. I was thinking of switching over to it, but have a question.

Do you want to still support the busybox crond ( aka allow users to switch to that one, by setting an explicit flag ) or remove it completely?

pixxon:
I switched over to the new scheduler. Users who want to opt into the old version can do so by specifying the `entrypoint` of the container to the old script.

I would assume that the old version will be deprecated anyway, it is only there for some grace period. So I did not document this alternative, but I can write something about it if you think it's worth.

m90:
> Users who want to opt into the old version can do so by specifying the entrypoint of the container to the old script.

I'd remove the old entrypoint. This patch should be able to support everything that has been supported previously, so there's no need to give users a choice. If there are things we cannot support that have been supported previously, we'd need to figure this out.

cmd/backup/main.go
```diff
@@ -65,4 +82,83 @@ func main() {
 	s.must(s.withLabeledCommands(lifecyclePhaseProcess, s.encryptArchive)())
 	s.must(s.withLabeledCommands(lifecyclePhaseCopy, s.copyArchive)())
 	s.must(s.withLabeledCommands(lifecyclePhasePrune, s.pruneBackups)())
+
+	return nil
+}
+
+func main() {
+	serve := flag.Bool("foreground", false, "run the backup in the foreground")
+	envFolder := flag.String("env-folder", "/etc/dockervolumebackup/conf.d", "location of environment files")
```
m90, line 91: Not sure if this needs to be configurable. Considering the binary will always be wrapped in a Docker image, we have a pretty clear idea of how the fs layout looks like, so I would think it's ok to hard code it.

Passing it in a flag would also all of a sudden open up a second way of configuring the tool, so if we really want it, it would need to be set in an env var.

I'd vote for keep using the hard coded value.

cmd/backup/main.go
```diff
@@ -65,4 +82,83 @@ func main() {
 	s.must(s.withLabeledCommands(lifecyclePhaseProcess, s.encryptArchive)())
 	s.must(s.withLabeledCommands(lifecyclePhaseCopy, s.copyArchive)())
 	s.must(s.withLabeledCommands(lifecyclePhasePrune, s.pruneBackups)())
+
+	return nil
+}
+
+func main() {
+	serve := flag.Bool("foreground", false, "run the backup in the foreground")
```
m90, line 90:
```suggestion
	serve := flag.Bool("foreground", false, "run the tool in the foreground")
```

"running the backup" is a bit ambiguous.

cmd/backup/main.go
```diff
@@ -65,4 +82,83 @@ func main() {
...
+		}
+
+		err = runBackup(c)
+		if err != nil {
+			logger.Info("unexpected error during backup", "error", err)
+			os.Exit(1)
+		}
+	}
+
+	os.Exit(0)
```
m90, line 163: I'd remove this as:
- it's the default behavior when there's nothing left to do for the program
- this can sneakily terminate the program while we still have dangling goroutines or similar running. Without force exiting, we'll notice such things quicker

cmd/backup/main.go
```diff
@@ -65,4 +82,83 @@ func main() {
...
+		logger.Info("interrupt arrived, stopping schedules")
+		ctx := cr.Stop()
+		<-ctx.Done()
+	} else {
+		logger.Info("executing one time backup")
+
+		c, err := loadEnvVars()
+		if err != nil {
+			logger.Info("could not load config from environment variables", "error", err)
+			os.Exit(1)
```
m90, line 153: I think we can improve that `func main` now has to manually exit in so many places. Previously, we had the deferred `recover` that would see all fatal errors and would be the only place where the code force exits the program.

I think we should introduce something similar here as well, i.e. another `must` function that is responsible for handling errors that happen in `func main`.

Maybe, it'd even be a good idea to start adding an `app` type that kind of works like `script`, but is responsible for managing collecting all dependencies of the application when running in the foreground.

m90 (commented):
I've left some comments, let me know what you think of them.

I'll also have to think about the error handling in `func main` and how it interacts with the config loading and constructing the `script` a bit more. Maybe this can be a bit more straightforward.

cmd/backup/configProvider.go
```diff
@@ -0,0 +1,89 @@
+// Copyright 2021-2022 - Offen Authors <hioffen@posteo.de>
```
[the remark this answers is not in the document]
pixxon: Done

Dockerfile
```diff
@@ -18,4 +18,4 @@ RUN apk add --no-cache ca-certificates
 COPY --from=builder /app/cmd/backup/backup /usr/bin/backup
 COPY --chmod=755 ./entrypoint.sh /root/
 
-ENTRYPOINT ["/root/entrypoint.sh"]
+ENTRYPOINT ["/usr/bin/backup", "--foreground"]
```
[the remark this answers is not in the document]
pixxon: Done

cmd/backup/configProvider.go
```diff
@@ -0,0 +1,89 @@
...
+		value, okValue := lookup(key)
+		location, okFile := lookup(key + "_FILE")
+
+		switch {
+		case okValue && !okFile: // only value
+			return value, true
+		case !okValue && okFile: // only file
+			contents, err := os.ReadFile(location)
+			if err != nil {
+				slog.Error(fmt.Sprintf("failed to read %s", location), "error", err)
```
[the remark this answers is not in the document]
pixxon: Where I would return an error, I changed it to `fmt.Errorf`. Only once case remained where I would need to signal error inside the cron function.

cmd/backup/configProvider.go
```diff
@@ -0,0 +1,89 @@
...
+			return "", false
+		default: // neither, ignore
+			return "", false
+		}
+	}
+
+	var c = &Config{}
+	if err := envconfig.Process("", c); err != nil {
+		slog.Error("failed to process configuration values", "error", err)
+		return nil, err
```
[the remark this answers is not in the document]
pixxon: Done

cmd/backup/configProvider.go
```diff
@@ -0,0 +1,89 @@
...
+	"fmt"
+	"log/slog"
+	"os"
+	"path/filepath"
+
+	"github.com/joho/godotenv"
+	"github.com/offen/envconfig"
+)
+
+type envProxy func(string) (string, bool)
```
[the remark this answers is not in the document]
pixxon: Done

cmd/backup/configProvider.go
```diff
@@ -0,0 +1,89 @@
...
+
+func loadEnvVars() (*Config, error) {
+	slog.Info("loading config from environment variables")
+	return loadConfig(os.LookupEnv)
+}
+
+func loadEnvFiles(folder string) ([]*Config, error) {
+	slog.Info(fmt.Sprintf("loading config from environment files from directory %s", folder))
+
+	items, err := os.ReadDir(folder)
```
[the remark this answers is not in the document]
pixxon: This is checked outside the call. The `loadEnvFiles` function tries to load from a location. If it does not exist, it will report that. It is up to the caller to decide if that error is recoverable.

cmd/backup/configProvider.go
```diff
@@ -0,0 +1,89 @@
...
+
+	return c, nil
+}
+
+func loadEnvVars() (*Config, error) {
+	slog.Info("loading config from environment variables")
+	return loadConfig(os.LookupEnv)
+}
+
+func loadEnvFiles(folder string) ([]*Config, error) {
```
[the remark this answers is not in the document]
pixxon: Done

cmd/backup/main.go
pixxon: I kept it configurable for ease of testing, but removed from final version.

cmd/backup/main.go
pixxon: Done

cmd/backup/main.go
pixxon: Done

cmd/backup/main.go
pixxon: For now I only created a wrapper `app` function that can return an `error`. I think it will be enough, there should be no panic between the recover handler in script and the main function.

cmd/backup/configProvider.go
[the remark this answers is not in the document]
m90: I understand that, but from reading the docs, I am not sure whether `os.ReadDir` will ever return such an error

> ReadDir reads the named directory, returning all its directory entries sorted by filename. If an error occurs reading the directory, ReadDir returns the entries it was able to read before the error, along with the error.

which is why I thought we'd need to cater for the case of an inexistent directory ourselve?.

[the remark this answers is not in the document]
pixxon: My understanding is that the folder not existing would certainly trigger an error during reading and hence it would be returned by `os.ReadDir`. It is also the behaviour that I observed.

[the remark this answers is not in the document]
m90: Ah, I did not know that `os.IsNotExist` also picks up syscall errors:

```
// IsNotExist returns a boolean indicating whether the error is known to
// report that a file or directory does not exist. It is satisfied by
// ErrNotExist as well as some syscall errors.
```

I was thinking it's just really checking for `os.ErrNotExist`. But in this case this should be working as is. Sorry for the confusion.

cmd/backup/configProvider.go
```diff
@@ -0,0 +1,89 @@
...
+	if err != nil {
+		slog.Error("failed to read files from env folder", "error", err)
+		return nil, err
+	}
+
+	var cs = make([]*Config, 0)
+	for _, item := range items {
+		if item.IsDir() {
+			slog.Info("skipping subdirectory")
+		} else {
```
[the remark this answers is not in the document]
m90: I was trying to say it could just be

```go
	for _, item := range items {
		if item.IsDir() {
			continue
		}
		p := filepath.Join(directory, item.Name())
		envFile, err := godotenv.Read(p)
		if err != nil {
			return nil, fmt.Errorf("error reading config file %s, error: %v", p, err)
		}
		lookup := func(key string) (string, bool) {
			val, ok := envFile[key]
			return val, ok
		}
		c, err := loadConfig(lookup)
		if err != nil {
			return nil, fmt.Errorf("error loading config from file %s, error: %v", p, err)
		}
		cs = append(cs, c)
	}
```

but it's ok as is now as well, no need to change anything.

cmd/backup/main.go
m90: I'll need some more focused time to look at this, so I'll have a look when I can and will feedback on this particular topic.

m90 (commented):
This looks very nice already. I left some minor comments, let me know if there are questions around them.

cmd/backup/main.go
```diff
@@ -4,21 +4,33 @@
...
 	"os"
+	"os/signal"
+	"syscall"
+
+	"github.com/robfig/cron/v3"
 )
 
-func main() {
-	s, err := newScript()
+func runBackup(c *Config) (ret error) {
```
m90, line 17:
```suggestion
func runScript(c *Config) (ret error) {
```

How about naming this `runScript` as it constructs a `script` object and then runs it?

cmd/backup/main.go
```diff
@@ -65,4 +82,82 @@ func main() {
...
+
+	return nil
+}
+
+func app(serve bool) error {
+	if serve {
+		cr := cron.New(
+			cron.WithParser(
+				cron.NewParser(
+					cron.SecondOptional | cron.Minute | cron.Hour | cron.Dom | cron.Month | cron.Dow | cron.Descriptor,
```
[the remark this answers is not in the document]
pixxon: Certainly, especially since it was referring to busybox. I added a nice table from the cron package with modifications to how we are parsing it.

cmd/backup/main.go
```diff
@@ -65,4 +82,82 @@ func main() {
 	s.must(s.withLabeledCommands(lifecyclePhaseProcess, s.encryptArchive)())
 	s.must(s.withLabeledCommands(lifecyclePhaseCopy, s.copyArchive)())
 	s.must(s.withLabeledCommands(lifecyclePhasePrune, s.pruneBackups)())
+
+	return nil
+}
+
+func app(serve bool) error {
```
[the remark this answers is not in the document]
pixxon: Done

cmd/backup/main.go
pixxon: Done

cmd/backup/main.go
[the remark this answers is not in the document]
m90: This turned out real good, thanks.

m90:
@pixxon One difference I noticed between the busybox cron and this version is that now, we'll happily register valid, but "impossible" cron schedules like "run on the 31st of Februrary".

Previously this would refuse to be added to the crontab with a message like

```
crond: user root: parse error at 31
```

In the end, the result is the same as the container runs, idling forever, but I am wondering if we should (if possible) also notify users who provided such an expression?

m90:
Ah, I just noticed you mentioned this in the documentation already. In any case, if there is a possibility to somehow make the cron library give us a warning about how a schedule will never run, I think we should log it. If not, I guess it's fine too.

m90:
Interesting, I found a terribly hacky way of doing this. If you parse the cron expression and ask for its next execution, the library call will panic in case the cron will never run. Expressions that will run pass this test just fine

```go
		sched, _ := cron.ParseStandard(config.BackupCronExpression)
		now := time.Now()
		next := sched.Next(now) // panics when the cron would never run
```

Not sure if I want to use this though.

(Stolen from here https://github.com/robfig/cron/issues/392#issuecomment-932586483)

PR #339, opened 2024-02-01
Exclusive file lock is released prematurely

m90:
It seems #263 broke the file lock as it wrapped the `unlock()` call in `must`, which means it would not get deferred anymore.


m90:
This is released in [v2.36.1](https://github.com/offen/docker-volume-backup/releases/tag/v2.36.1)

PR #347, opened 2024-02-06
Move cron scheduling inside application

### LINKED ISSUE #268: Consider using a dedicated cron package

m90:
Closes #268


m90:
@pixxon Just to give un update on the status of this: I did some minor tweaking in regards to error handling and logging and would think this is now ready to go from a code perspective.

As we're now having a long running process though, I'd still like to confirm this doesn't leak memory somehow, so I'll let the current state of this branch (`v2.37.0-alpha.3`) run on all of my setups and check memory usage over the course of a week. If this looks unsuspicous, I'd say let's ship it.

pixxon:
Sounds great!

I am not familiar with golang, but maybe there are some tools to help finding memory leaks. ( In the c++ word, there are some great static analyzers that help, or even better are the memory sanitizers that check for addresses that were not freed properly. )

m90:
The only way to really leak memory in Go is to have goroutines running that never return, meaning everything in their scope isn't possible to be garbage collected. I definitely managed to do this before, even when having static analysis look for such problems. I don't think it's likely here, but I'd feel safer if I've seen it running for a bit.

One simple thing we could do is print the numbers of Go routines from time to time to see if the number stays stable, but I guess I'll defer that until it looks like there really is a leak.

m90:
I got nerd sniped and fired up [`pprof`](https://pkg.go.dev/net/http/pprof) which let me indeed finf a leak: the Docker client leaks a persistent HTTP connection when you don't call `Close` on the client object.

After doing this, the number of Goroutines seems stable when I run a per-minute schedule locally for ~15mins or so.

See 68ba6ce5a1c4523bde3035e415f7396bd90d54be as a fix

I'll push this as `v2.37.0-alpha.4` and deploy it.

m90:
Seems the advice on closing it is right there in the docs https://pkg.go.dev/github.com/docker/docker/client#section-readme ....

PR #358, opened 2024-02-13
Env vars should propagate when using `conf.d`

### LINKED ISSUE #357: Env vars stopped propagating when using `conf.d`
Starting with `v2.37.0`, environment variables set on the `backup` container will not propagate to backup runs defined by a file in `conf.d` anymore. This is a bug and needs to be fixed.

If you are affected by this, the current solution would be to downgrade to `v2.36.1` and wait until this issue is closed.

m90:
Closes #357

### DIFF
--- a/cmd/backup/config_provider.go
+++ b/cmd/backup/config_provider.go
@@ -49,8 +49,9 @@ func loadEnvVars() (*Config, error) {
 }
 
 type configFile struct {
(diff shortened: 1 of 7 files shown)


m90:
This can use some further refactoring, but I'll merge and release it now in order to unbreak the latest version.

PR #361, opened 2024-02-13
add OCI labels, rework tagging

pxlfrk:
i'm using [Renovate](https://github.com/renovatebot/renovate) to automatically update dependencies. Renovate (and similar tools as well) support the docker label `org.opencontainers.image.source`, also known as [OCI](https://github.com/opencontainers/image-spec/blob/main/annotations.md) labels to point to sources and add the release notes in the created PRs --> see for example a PR for [tandoor](https://github.com/TandoorRecipes/recipes):

image

I noticed, that this docker-volume-backup currently doesn't leverage these labels and implemented them.
When possible i tried to stick to variables instead of hardcoding informations to keep the package dynamically, in case someone forks and wants to distribute it by himself.

Furthermore i kind of reworked the tagging code and replaced it with the popolur github-action [metadata-action](https://github.com/marketplace/actions/docker-metadata-action).

I stuck with the current logic and implemented a rule, that e.g. the tag `v2.15.4-alpha.2` doens't get the `latest`-tag.

```
# tag releases as latest, except prerelease tags like `v2.0.0-alpha.x`
type=raw,value=latest,enable=${{ !contains(github.ref, 'alpha') }}
```

See the [full documentation](https://github.com/docker/metadata-action?tab=readme-ov-file#typesemver) for the semver-mapping.

I look forward to improvements, reviews and tips!

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -18,7 +18,6 @@ jobs:
       - name: set Environment Variables
         id: env
         run: |
-          version_tag="${{github.ref_name}}"
           echo "NOW=$(date +'%F %Z %T')" >> $GITHUB_ENV
 
       - name: Docker meta
@@ -32,16 +31,10 @@ jobs:
           # define global behaviour for tags
           flavor: |
             latest=false
-          # generate Docker tags based on the following events/attributes
+          # specify one tag which never gets set, to prevent the tag-attribute being empty, as it will fallback to a default
           tags: |
-            # tag releases as latest, except prerelease tags like `v2.0.0-alpha.x`
-            type=raw,value=latest,enable=${{ !contains(github.ref, 'alpha') }}
             # output v2.42.1-alpha.1 (incl. pre-releases)
-            type=semver,pattern=v{{version}}
-            # output v2.42.1 (excl. pre-releases)
-            type=semver,pattern=v{{major}}.{{minor}}.{{patch}}
-            # output v2 
-            type=semver,pattern=v{{major}}
+            type=semver,pattern=v{{version}},enable=false
           labels: |
             org.opencontainers.image.title=${{github.event.repository.name}}
             org.opencontainers.image.description="Backup Docker volumes locally or to any S3, WebDAV, Azure Blob Storage, Dropbox or SSH compatible storage"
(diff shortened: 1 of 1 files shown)


m90:
Thanks for opening this PR. I am happy to add the labels, but I would prefer to stick with the current state of the tagging code. It took me ages to get this going and I just could not understand how the metadata action is expecting me to configure it, which is why the custom script is there. As I'll have to keep maintaining this I'd like to keep what I understand. Or does your patch fix anything that is broken in this regard?

As far as I understand we can add the labels without changing the tagging logic, right?

pxlfrk:
Thanks! No, my patch doesn't fix somethink broken, after looking into this and reading the manuals i just found this solution some kind of easier and more streamlined. Since i haven't run the workflow, i couldn't test this fix yet and i don't know how to do this without setting up the workflow in my fork as well.

Until i figured this out and tested this, i'll change the PR, so that the action only adds the labels, and the tag will be provided by your current workflow-step.

m90:
Thanks!

m90:
I just released an image using this for the first time and when inspecting the image, some values are double quoted

```json
            "Labels": {
                "org.opencontainers.image.created": "2024-02-15 UTC 11:05:49",
                "org.opencontainers.image.description": "\"Backup Docker volumes locally or to any S3, WebDAV, Azure Blob Storage, Dropbox or SSH compatible storage\"",
                "org.opencontainers.image.documentation": "\"https://offen.github.io/docker-volume-backup/\"",
                "org.opencontainers.image.licenses": "\"MPL-2.0\"",
                "org.opencontainers.image.revision": "a01fc3df3f7030569661e324687d6ae97bc84d1d",
                "org.opencontainers.image.source": "https://github.com/offen/docker-volume-backup",
                "org.opencontainers.image.title": "docker-volume-backup",
                "org.opencontainers.image.url": "\"https://offen.github.io/docker-volume-backup/\"",
                "org.opencontainers.image.vendor": "offen",
                "org.opencontainers.image.version": "v2.37.3"
            }
```

Is this expected or should we do something about it?

m90:
Seems the docs are using unqoted literal strings https://github.com/docker/metadata-action?tab=readme-ov-file#inputs

m90:
Already did that, thanks https://github.com/offen/docker-volume-backup/pull/365

pxlfrk:
It works, thanks! Now i cann see the changelog directly in the PRs :)
Just wanted to let you (& possible fellow readers of this PR) know ... 

image

PR #362, opened 2024-02-13
Conditionally run docs-pages-build

pxlfrk:
This PR changes the workflow to rebuild the docs on github-pages to only run, when content inside the `docs/` folder is changed. Currently with every push on main the docs are rebuilt, even when no changes have been made.

### DIFF
--- a/.github/workflows/deploy-docs.yml
+++ b/.github/workflows/deploy-docs.yml
@@ -3,6 +3,7 @@ name: Deploy Documenation site to GitHub Pages
 on:
   push:
     branches: ['main']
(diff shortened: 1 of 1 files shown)


m90 (approved):
Makes sense, thank you.

m90:
I just noticed (obviously _after_ merging) we forgot one thing here: the docs should obviously also be deployed anew when the workflow file itself changes, i.e. something in the build process changes. No need to fix it right now though, I'll put it on my list.

m90:
Fixed in fb4663b087e3897026e048594616f4b78fb48f11

PR #417, opened 2024-05-10
backupp all

jattento:


cmd/backup/stop_restart.go
```diff
@@ -129,12 +129,7 @@ func (s *script) stopContainersAndServices() (func() error, error) {
 	if err != nil {
 		return noop, errwrap.Wrap(err, "error querying for containers")
 	}
-	containersToStop, err := s.cli.ContainerList(context.Background(), container.ListOptions{
-		Filters: filters.NewArgs(filters.KeyValuePair{
-			Key:   "label",
-			Value: filterMatchLabel,
-		}),
-	})
+	containersToStop, err := s.cli.ContainerList(context.Background(), container.ListOptions{All: true})
```
m90, line 132: Not sure about your plans, but this will also mean that if you run this program in a container, the container will also stop itself, which might not be what you want.

jattento: @m90 Thanks for letting me know, going to try to find a workaround.

I want to backup some dynamically created docker containers which I have no way to add tags (nextcloud AIO) and I need them to stop during backup

PR #456, opened 2024-08-09
Feature: PGP Asymmetric Encryption

lbsekr:
While this project already provides symmetric pgp encryption, it lacks a way to encrypt backups asymmetric.
Therefore this pull request provides that functionality.

Therefore the `encryptArchive` function has been changed. All changes have been tested and documented.

I have not done much programming in Golang in the last year. Suggestions for improvement are welcome.


docs/reference/index.md
```diff
@@ -332,10 +332,19 @@ You can populate below template according to your requirements and use it as you
...
 ########### BACKUP ENCRYPTION
 
-# Backups can be encrypted using gpg in case a passphrase is given.
+# Backups can be encrypted symmetrically using gpg in case a passphrase is given.
 
 # GPG_PASSPHRASE="<xxx>"
 
+# Backups can be encrypted asymmetrically using gpg in case publickeys are given.
+
+# GPG_PASSPHRASE= |
```
m90, line 341:
```suggestion
# GPG_PUBLIC_KEYS= | 
```

cmd/backup/encrypt_archive.go
```diff
@@ -4,20 +4,87 @@
...
 	"path"
 
+	"github.com/ProtonMail/go-crypto/openpgp/armor"
 	openpgp "github.com/ProtonMail/go-crypto/openpgp/v2"
 	"github.com/offen/docker-volume-backup/internal/errwrap"
 )
 
-// encryptArchive encrypts the backup file using PGP and the configured passphrase.
-// In case no passphrase is given it returns early, leaving the backup file
+func readArmoredKeys(data []byte) (openpgp.EntityList, error) {
```
m90, line 19: I'm not entirely sure this function is needed. It could be refactored to 
```go
func readArmoredKeys(data []byte) (openpgp.EntityList, error) {
    return openpgp.ReadArmoredKeyRing(bytes.NewReader(data))
}
```

which means we could also just call this inline further down?

cmd/backup/encrypt_archive.go
```diff
@@ -4,20 +4,87 @@
...
+	if err != nil {
+		return nil, nil, errwrap.Wrap(err, "error preparing encryption")
+	}
+
+	_, name := path.Split(s.file)
+	dst, err := openpgp.Encrypt(armoredWriter, entityList, nil, nil, &openpgp.FileHints{
+		FileName: name,
+	}, nil)
+	if err != nil {
+		return nil, nil, err
```
m90, line 47: This error should still be wrapped like all the others.

cmd/backup/encrypt_archive.go
```diff
@@ -4,20 +4,87 @@
...
+	if s.c.GpgPassphrase == "" {
+		return nil, nil, nil
+	}
+
+	_, name := path.Split(s.file)
+	dst, err := openpgp.SymmetricallyEncrypt(outFile, []byte(s.c.GpgPassphrase), &openpgp.FileHints{
+		FileName: name,
+	}, nil)
+	if err != nil {
+		return nil, nil, err
```
m90, line 67: This error should still be wrapped.

cmd/backup/encrypt_archive.go
```diff
@@ -4,20 +4,87 @@
...
+		dst.Close()
+		armoredWriter.Close()
+		return nil
+	}, err
+}
+
+func (s *script) encryptSymmetrically(outFile *os.File) (io.WriteCloser, func() error, error) {
+	if s.c.GpgPassphrase == "" {
+		return nil, nil, nil
+	}
```
m90, line 60: As you now introduced checking for proper config in `encryptArchive`, I'm not sure this check is still needed. Is there still a way 'encryptSymmetrically` could be called _without_ a passphrase being set?

The same applies to the check in `encryptAsymmetrically`.

cmd/backup/config.go
```diff
@@ -47,6 +47,7 @@ type Config struct {
 	BackupExcludeRegexp           RegexpDecoder   `split_words:"true"`
 	BackupSkipBackendsFromPrune   []string        `split_words:"true"`
 	GpgPassphrase                 string          `split_words:"true"`
+	GpgPublicKeys                 string          `split_words:"true"`
```
m90, line 50: I'm a bit torn on the plural name. Other keys in the config that use a plural s are comma separated list. so on first read I expected this to be a comma separated list as well, which it isn't.

Maybe this could be called `GpgPublicKeyRing` or `GpgArmoredKeyRing`, picking up the terminology from the gpg package?

cmd/backup/encrypt_archive.go
```diff
@@ -4,20 +4,87 @@
...
+// In case no passphrase or publickey is given it returns early, leaving the backup file
 // untouched.
 func (s *script) encryptArchive() error {
-	if s.c.GpgPassphrase == "" {
+
+	var encrypt func(outFile *os.File) (io.WriteCloser, func() error, error)
+
+	switch {
+	case s.c.GpgPassphrase != "" && s.c.GpgPublicKeys != "":
+		return errors.New("error in selecting asymmetric and symmetric encryption methods: conflicting env vars are set")
```
m90, line 82:
```suggestion
		return errwrap.Wrap(nil, "error in selecting asymmetric and symmetric encryption methods: conflicting env vars are set")
```
This is a bit non-obvious,but like this the error will start recording a poor man's stack trace for us.

m90 (commented):
Thanks for this valuable addition. I left a few comments inline, nothing major. Let me know what you think.

cmd/backup/encrypt_archive.go
lbsekr: yeah, you are right. i probably overlooked while refactoring, I will fix this

cmd/backup/encrypt_archive.go
lbsekr: I decided not to wrap it here and do it in the caller func instead. So the warping is the same for both at a single point.

cmd/backup/config.go
lbsekr: Yes, it can be a bit misleading and does not fit well into this project's terminology. I think using `GpgPublicKeyRing` is best, so its clear public keys are expected.

cmd/backup/encrypt_archive.go
lbsekr: yes, again orphaned code. I will remove it

cmd/backup/encrypt_archive.go
lbsekr: works for me

cmd/backup/encrypt_archive.go
lbsekr: fixed in [96f996b](https://github.com/offen/docker-volume-backup/pull/456/commits/96f996b25947c66197177ee38823094b61fe98e2)

cmd/backup/encrypt_archive.go
lbsekr: fixed in [3b96590](https://github.com/offen/docker-volume-backup/pull/456/commits/3b96590c9880cb4c38a9201551197ecea1522dc0)

cmd/backup/config.go
lbsekr: fixed in [c76c13d](https://github.com/offen/docker-volume-backup/pull/456/commits/c76c13d02ee28226e926aba8e8c1b7101187a584)

test/gpg-asym/docker-compose.yml
```diff
@@ -0,0 +1,37 @@
...
+  backup:
+    image: offen/docker-volume-backup:${TEST_VERSION:-canary}
+    restart: always
+    environment:
+      BACKUP_CRON_EXPRESSION: 0 0 5 31 2 ?
+      BACKUP_FILENAME: test.tar.gz
+      BACKUP_LATEST_SYMLINK: test-latest.tar.gz.gpg
+      BACKUP_RETENTION_DAYS: ${BACKUP_RETENTION_DAYS:-7}
+      GPG_PUBLIC_KEY_RING: | 
+        -----BEGIN PGP PUBLIC KEY BLOCK-----
```
m90, line 11: I wonder if it would make sense to remove the duplication and instead pass the asc-file using GPG_PUBLIC_KEY_RING_FILE, which I would assume is what most people would be using in the wild as well. What do you think?

test/gpg-asym/test-priv.asc
m90, line 1: The ssh test currently generates a keypair for every run and does not check in any test keys. Just so you know this is an option too https://github.com/offen/docker-volume-backup/blob/336e12f8747c80a25a49e639416e3f5c1120afa8/test/ssh/run.sh#L11

I'm fine with either choice.

test/gpg-asym/run.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/sh
+
+set -e
+
+cd "$(dirname "$0")"
+. ../util.sh
+current_test=$(basename $(pwd))
+
+export LOCAL_DIR=$(mktemp -d)
+gpg-agent --daemon
```
m90, line 10: Just out of curiosity: why is this needed?

cmd/backup/encrypt_archive.go
```diff
@@ -4,20 +4,72 @@
...
+	dst, err := openpgp.Encrypt(armoredWriter, entityList, nil, nil, &openpgp.FileHints{
+		FileName: name,
+	}, nil)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	return dst, func() error {
+		dst.Close()
+		armoredWriter.Close()
```
m90, line 40: These two function calls can return an error which needs to be checked. This is why CI is currently red.

test/gpg-asym/docker-compose.yml
lbsekr: What do you mean with remove the duplication. `GPG_PUBLIC_KEY_RING` and `-----BEGIN PGP PUBLIC KEY BLOCK`. Tbh, parsing a file would probably be cleaner, since putting the keyring in the compose.yaml makes the file unnecessarily large and hard to read.

Would you prefer to remove the `GPG_PUBLIC_KEY_RING` option altogether and introduce the `GPG_PUBLIC_KEY_RING_FILE` or a tandem solution. im fine with both but would prefer a standalone file solution.

cmd/backup/encrypt_archive.go
lbsekr: I will add error-handling

test/gpg-asym/run.sh
lbsekr: Decrypting fails without it. But I think this can be avoided when adjusting the decryption cli command. I will look into it

lbsekr: specifically, it is used for subsequent requests. e.g. importing the private key and then decrypting with it.

test/gpg-asym/docker-compose.yml
m90: Sorry for not being clear. All configuration values can also be read from a file if you supply a location using `CONFIG_KEY_FILE`, so this already works out of the box, you can pass a full key using `GPG_PUBLIC_KEY_RING` or the location to a key file using `GPG_PUBLIC_KEY_RING_FILE`.

See the topmost note here https://offen.github.io/docker-volume-backup/reference/

lbsekr: Nice. I will then pass the file in the compse.yaml

cmd/backup/encrypt_archive.go
lbsekr: fixed in [5cfa74c](https://github.com/offen/docker-volume-backup/pull/456/commits/5cfa74cb1bb6493e18e48f64444a0eef13760426)

test/gpg-asym/test-priv.asc
lbsekr: i implemented generic key generation for testing in [e44fd35](https://github.com/offen/docker-volume-backup/pull/456/commits/e44fd352e9d1383115cd917e47f8e32e39207219)

test/gpg-asym/docker-compose.yml
lbsekr: done, [e44fd35](https://github.com/offen/docker-volume-backup/pull/456/commits/e44fd352e9d1383115cd917e47f8e32e39207219)

test/gpg-asym/run.sh
lbsekr: While explicitly starting it is not required, the gpg agent is generally required in my understanding.. [1407661](https://github.com/offen/docker-volume-backup/pull/456/commits/14076614091ef0724756706d3e09ea363e1f2996)

lbsekr:
thanks for your reviews. I just hope all this is getting squashed 🙂

cmd/backup/encrypt_archive.go
```diff
@@ -38,14 +91,11 @@ func (s *script) encryptArchive() error {
...
-	_, name := path.Split(s.file)
-	dst, err := openpgp.SymmetricallyEncrypt(outFile, []byte(s.c.GpgPassphrase), &openpgp.FileHints{
-		FileName: name,
-	}, nil)
+	dst, closeCallback, err := encrypt(outFile)
 	if err != nil {
 		return errwrap.Wrap(err, "error encrypting backup file")
 	}
-	defer dst.Close()
+	defer closeCallback()
```
m90, line 98: This will now also need its error being handled. Check https://trstringer.com/golang-deferred-function-error-handling/ for an explanation on how to do this for a deferred function call.

m90:
This looks good to me now. There's a tiny complaint by the linter still, but once that's fixed I'd say this is good to go.

cmd/backup/encrypt_archive.go
lbsekr: should line 92, 104 also be handled

m90: Ideally, yes. I think the linter does not complain here as it's not part of the changeset. If you still want to add this, it'd be great.

lbsekr: No problem, I looked through this repo and its denfiently something that needs to be done in other files as well. I will do it in `encrypt_archive.go`

lbsekr: fixed in [007b0eb](https://github.com/offen/docker-volume-backup/pull/456/commits/007b0eb5b222eb148b1d512a358cebb6a5516a93)

m90 (approved):
Thanks for the feature and the conversation, that's much appreciated. I'll (squash) merge this when #457 is ready so I can cut a release for these changes then. I'll let you know when that happens 🎩

m90:
This is now included in [v2.42.0](https://github.com/offen/docker-volume-backup/releases/tag/v2.42.0)

PR #457, opened 2024-08-09
Add "none" compression type

### LINKED ISSUE #455: [Feature] Allow BACKUP_COMPRESSION to be disabled
I'd like to be able to create uncompressed backup archives as plain old `.tar`s. It looks like [getCompressionWriter](https://github.com/offen/docker-volume-backup/blob/main/cmd/backup/archive.go#L94) doesn't support disabling compression and the app always [defaults to gzip compression](https://github.com/offen/docker-volume-backup/blob/main/cmd/backup/config.go#L32).

jzebedee:
Fixes #455


cmd/backup/archive.go
```diff
@@ -83,16 +83,20 @@ func compress(paths []string, outFilePath, algo string, concurrency int) error {
 		return errwrap.Wrap(err, "error closing compression writer")
 	}
 
-	err = file.Close()
-	if err != nil {
-		return errwrap.Wrap(err, "error closing file")
+	if file != compressWriter {
```
m90, line 86: I find it a bit confusing to skip the closing here (I do understand the closing has already happened above and we shouldn't close the file again), but from reading the code right now I don't really understand that this can be a possibility. How about skipping the closing in the condition above instead which hopefully makes it a bit more obvious:

```go
	if file != compressWriter {
		err = compressWriter.Close()
		if err != nil {
			return errwrap.Wrap(err, "error closing compression writer")
		}
	}

	err = file.Close()
	if err != nil {
		return errwrap.Wrap(err, "error closing file")
	}
```

m90 (commented):
Could you also add the new option to the documentation? Thanks,

cmd/backup/archive.go
m90: Alternatively, instead of leaking a detail from `getCompressionWriter` to its callers, you could also wrap `file` in a `WriteCloser` that just passes through writes to the file and does nothing on close:

```go
type passThroughWriteCloser struct {
	target io.WriteCloser
}

func (p *passThroughWriteCloser) Write(b []byte) (int, error) {
	return p.target.Write(b)
}

func (p *passThroughWriteCloser) Close() error {
	return nil
}
```
and
```go
	case "none":
		return &passThroughWriteCloser{file}, nil
```
Like this, consumers don't need to know about special cases in the returned object. Maybe this is the better option, what do you think?

m90:
Thanks, I still added a test, let's merge it.

PR #531, opened 2025-02-06
Passing a full duration as retention period is more flexible

### LINKED ISSUE #528: Retention improvement
why is the retention mechanism not more flexible, for example, to allow setting a parameter in days, hours, and minutes? I had a situation where I needed to make a backup every hour and store only the last two archives, but I could not achieve this with the existing mechanisms

m90:
Closes #528


m90:
Closing this while #528 is going stale

PR #538, opened 2025-02-21
Update stop-containers-during-backup.md

jastlw:
Read-only for /var/run/docker.sock:/var/run/docker.sock is wrong. Otherwise, a stop of different containers may not be possible.

Here is it all right: https://offen.github.io/docker-volume-backup/recipes/#using-mariadb-dumpmysqldump-to-prepare-the-backup

### DIFF
--- a/docs/how-tos/stop-containers-during-backup.md
+++ b/docs/how-tos/stop-containers-during-backup.md
@@ -31,7 +31,7 @@ services:
       BACKUP_STOP_DURING_BACKUP_LABEL: service1
     volumes:
       - data:/backup/my-app-backup:ro
-      - /var/run/docker.sock:/var/run/docker.sock:ro
+      - /var/run/docker.sock:/var/run/docker.sock
 
 volumes:
   data:


m90:
Thanks for catching this, would you mind also fixing this mistake in the main README file while we're at it?

m90:
Could you share further details about the setup where mounting the socket in readonly mode creates failures for you? I just tried running the tests using `:ro` for the socket and restarting containers and scaling services still worked as expected.

m90:
I migrated the entire test suite to use `:ro` mounts for sockets in #546 and everything passes as expected, which means there must be something special about your setup. Can you share a. the configuration you use and b. the error you are encountering?

jastlw:
@m90 Thanks for testing and all the work you have done. 😍 

I think you are right: I want to execute the `docker-volume-backup`  from a remote machine. Therefore I bound the docker socket over sshfs. There were some faults.
Sorry for the trouble. 

You could think about to also make it here only ro: https://offen.github.io/docker-volume-backup/recipes/#using-mariadb-dumpmysqldump-to-prepare-the-backup

m90:
Updated the docs to always use `:ro` in d8ac5ae7e6e61116ae6000d9a5b2313e63b56545

Issue #471, opened 2024-09-13
Would it be possible to support Google Drive


dustin-auby:
Hi, would it be possible to use google drive or event a google storage bucket. we only use google services and at the moment have no way to implement this.

m90:
Google Cloud Storage does support the S3 protocol, so you can definitely use it as a backend for this tool using the S3 configuration: https://cloud.google.com/storage/docs/interoperability

As for Google Drive, I'd think it's probably possible but also complicated to an extent where I'm not sure it's worth it. This has also been discussed here #103

luizguilhermesj:
I created a script for Google Drive upload and retention (totally inspired by docker-volume-backup) 
I pushed it into an image on:
https://hub.docker.com/r/luizguilhermesj/docker-volume-google-drive-backup

I created it this afternoon using my free quota of http://cursor.com/
I have no previous experience with go, so don't shame me please 😅 

context: I got lockedout of my azure account, so I needed to rush into a solution to backup my stuff elsewhere and I do have a lot of space on Google Drive.
So the script doesn't even have tests, I'll try to improve it moving forward.
And I do see the pain it would be to support an integration with it.

m90:
@luizguilhermesj if you feel like wrangling cursor into writing a Google Drive backend that can be used in addition to what already exists, I'd be happy to merge it. From a brief look at the code it looks it should be possible.

PR #613, opened 2025-07-20
feat: Add Google Drive storage backend with custom endpoint support

luizguilhermesj:
Relates to #471

- Implement Google Drive storage backend using service account authentication
- Add support for custom Google Drive API endpoints for testing
- Include comprehensive test setup with mock services
- Add configuration options for Google Drive credentials and folder ID
- Update documentation with Google Drive configuration examples


docs/reference/index.md
```diff
@@ -386,6 +386,45 @@ The values for each key currently match its default.
 
 # DROPBOX_REFRESH_TOKEN=""
 
+########### GOOGLE DRIVE STORAGE
+
+# The JSON credentials for a Google service account with access to Google Drive. The value should be the path to a json file.
```
m90, line 391: This is correct, but judging from the issues I've seen raised in this repo over time, I think it wouldn't hurt to also mention this file needs to be mounted into the container.

cmd/backup/script.go
```diff
@@ -225,6 +227,21 @@ func (s *script) init() error {
 		s.storages = append(s.storages, dropboxBackend)
 	}
 
+	if s.c.GoogleDriveCredentialsJSON != "" {
+		googleDriveConfig := googledrive.Config{
+			CredentialsJSON:    s.c.GoogleDriveCredentialsJSON,
```
m90, line 232: Is it also possible to pass the JSON as a string instead of as a file location here? In case yes, we could leverage the existing configuration mechanism so that:

1. People can pass a JSON string to `GOOGLE_DRIVE_CREDENTIALS_JSON` if they don't want to / can't mount the file
2. People can pass a file location using `GOOGLE_DRIVE_CREDENTIALS_JSON_FILE` which is automatically provided by the  current configuration package

test/googledrive/run.sh
```diff
@@ -0,0 +1,81 @@
+#!/bin/sh
+
+set -e
+
+echo "DEBUG: Script started"
```
m90, line 5: It'd be good to remove all the logging from here once this is done, but I guess that's what you planned anyways.

cmd/backup/script.go
luizguilhermesj: I was trying to aim to only accept FILE, as it is an oddly long string to be putting on an envvar, but gemini got lost at some point and added it for me.
Also, I know _FILE is a pattern everywhere, but I have not seen it on other variables here, should we add it?

Let me know if you think just the JSON is enough, or if you prefer both so I an either remove that logic or fix the envvars and docs.

luizguilhermesj:
> Thanks a lot for following up on this so quickly. Two questions before I have a closer look:
> 
> 1. the integration test you added is failing, it seems pruning is either broken or expectations are wrong, do you have any idea why?
> 2. The linter left a tiny complaint about using a deprecated package, let me know if you need any help with fixing that

The tests were mainly copied from dropbox. But they were anchored by messages from storage.go, so I think it would be better to decouple it and fail with any ERROR message.
So I simplified it a lit bit and I think it is reporting both errors and success more [clearly now](https://github.com/user-attachments/assets/15ce3e77-06be-4149-b19d-bc3d23b2b341).
Plese let me know if there`s a better way or if it should respect the previous pattern.

I'm not sure if its a pattern, but the `set -e` makes the tests pass without errors, despite `logs=$(docker compose exec backup backup)` failing.
So I [wrapped that command](https://github.com/offen/docker-volume-backup/pull/613/files/78a0fc70a26fca775a5f275073338cdc52bb47d7#diff-59d9fe3f9b49400d50fa60fbed397902d809f6a7f8e46d7fd4d9a5e946949523R25-R28) with `set +e` and `set -e` to output the error to the log make it fail when it should, WDYT?

For the lint issue, I was about to ask for your help, but then I asked Gemini and he did [this](https://github.com/offen/docker-volume-backup/pull/613/commits/8ba12b0e7a4ec92606702cfdfce6f9d9745c7870).
The tests are still passing, so I think it is correct? 😅

cmd/backup/script.go
m90: > Also, I know _FILE is a pattern everywhere, but I have not seen it on other variables here, should we add it?

`_FILE` is supported automatically. If you define `FOO_BAR`, `FOO_BAR_FILE` will be accepted as well. I would also think passing the entire JSON blob as an env var will rarely be done, but like this, it'd be consistent with what's already there.

m90:
> So I [wrapped that command](https://github.com/offen/docker-volume-backup/pull/613/files/78a0fc70a26fca775a5f275073338cdc52bb47d7#diff-59d9fe3f9b49400d50fa60fbed397902d809f6a7f8e46d7fd4d9a5e946949523R25-R28) with set +e and set -e to output the error to the log make it fail when it should, WDYT?

This must mean that now the `backup` command exits non-zero in this scenario, which is unwanted (hence, the Dropbox test doesn't need it). I'll need to look into the code why that happens.

> For the lint issue, I was about to ask for your help, but then I asked Gemini and he did [this](https://github.com/offen/docker-volume-backup/pull/613/commits/8ba12b0e7a4ec92606702cfdfce6f9d9745c7870).

Exactly this. Thanks,

luizguilhermesj:
hey @m90 thanks for your kind review.
ILet me take a better look at it during the weekend (when I have more time) and I'll pay closer attention to what the AI is doing so I can stop pushing silly mistakes like that 😅

luizguilhermesj:
hey @m90, comming back to it =]

> > So I [wrapped that command](https://github.com/offen/docker-volume-backup/pull/613/files/78a0fc70a26fca775a5f275073338cdc52bb47d7#diff-59d9fe3f9b49400d50fa60fbed397902d809f6a7f8e46d7fd4d9a5e946949523R25-R28) with set +e and set -e to output the error to the log make it fail when it should, WDYT?
> 
> This must mean that now the `backup` command exits non-zero in this scenario, which is unwanted (hence, the Dropbox test doesn't need it). I'll need to look into the code why that happens.

I took a better look at dropbox, and it seems like it is the same behavior =/

1. I've changed the `user_v2.yaml` there so the response for  creating a folder would fail: [screenshot](https://github.com/user-attachments/assets/927e15d8-9315-49f6-ae49-f47f4710af90)
2. Then I run the tests: [screenshot](https://github.com/user-attachments/assets/e585511a-59cf-4f0f-b0f5-5bf13bbbbbb5)
3. Then I added `set +e` and re-run: [screenshot](https://github.com/user-attachments/assets/50cea567-8c5c-4c8c-94b5-df706b3f2a78)

Without further looking into code, I guess we don't want non-zero exit codes in order to prevent the container from dying due to one backup failure, so the schedule (and possibly other backends) can carry on, am I right?

With that guess, I figured it would be good to test this failure on a normal run as well, <ins>**but it didn't fail**</ins>, it actually behaved like I was expecting the tests to (keeping the container alive and displaying the error message): [screenshot](https://github.com/user-attachments/assets/2e651fb5-2910-44f0-8d7e-89e725c0b42b)

I have yet to learn how to properly debug golang, so my hope is that you can easily spot what is going on 😅
My better guess would be that there's something in the normal flow preventing exit codes from overflowing to the container, but in the test env it is either not implemented or not working properly.

Any ideas?

I didn't have much time this weekend 😢 so I couldn't look into the other 2 issues, but: 
- _FILE:
  - It seems easy, I'll do it as soon as possible
- Tests:
  - The AI generated a very dummy mock api that doesn't really do anything, therefore the tests are also not testing anything. I'll take some time (next weekend) to think what test cases would be useful here and how to properly structure them.

m90:
I think the Dropbox tests work as expected. When querying for what files exist, the mock will always return two existing files, which are backdated by these lines:

```
sed -i 's/SERVER_MODIFIED_1/'"$(date "+%Y-%m-%dT%H:%M:%SZ")/g" $SPEC_FILE
sed -i 's/SERVER_MODIFIED_2/'"$(date "+%Y-%m-%dT%H:%M:%SZ" -d "14 days ago")/g" $SPEC_FILE
```

so the tests go kind of like this

- Create an backup and test if this exits cleanly, pruning is not considered
- Query the mock Dropbox for existing files, receive 2 files, everything would be pruned, which is not done, printing a warning. Assert that the warning is there
- Query the mock Dropbox for existing files, receive 2 files again, one of the old enough to be pruned, print the expected line. Assert that the corresponding log line is printed

When I change the mock like you shown in the screenshot, I get a failing test (exits 1), which is expected.

> My best guess would be that there's something in the normal flow preventing exit codes from overflowing to the container, but in the test env it is either not implemented or not working properly.

Since all test cases use `set -e`, failures propagate correctly. There are also test cases that check for non-zero exit codes, calling `set +e` beforehand (e.g. `collision`). Are you sure you aren't confusing what `set -e` and `set +e` do?

---

Even if it's not perfect, I think we could get the same done for Google Drive by adding a proper response here: https://github.com/offen/docker-volume-backup/pull/613/files#diff-3af2262cc057a6c13dfc4c4b0924326705ddfee2a56a7492bc7690194437dc73R95-R119

luizguilhermesj:
>When I change the mock like you shown in the screenshot, I get a failing test (exits 1), which is expected.

Sorry, I just checked and it does exit with [code 1](https://github.com/user-attachments/assets/0b0c842d-0d83-4f59-8ab0-47ca82b49d7c).

I assumed it didn't because there's no output.
Why do we have output on a normal run but not on tests?
Is there a reason for that?
(please forgive my ignorance, I'm a front-end guy with poor knowledge on shell scripts and ZERO on golang)

But anyways, I was stuck on this thinking I was doing something wrong, but turns out that GoogleDrive behaves the same.
So if it is expected, we are all good 🎉 

>Are you sure you aren't confusing what set -e and set +e do?

Bro, I had to google it to understand what it does and I'm not conpletly sure I got it. I just reported what I did with the help of AI 😅  which was to "wrap" the backup exec with it, [like this](https://github.com/user-attachments/assets/713db154-7d0a-4069-ae1a-4ef3673bd371).

>Even if it's not perfect, I think we could get the same done for Google Drive by adding a proper response here: https://github.com/offen/docker-volume-backup/pull/613/files#diff-3af2262cc057a6c13dfc4c4b0924326705ddfee2a56a7492bc7690194437dc73R95-R119

That's the plan!
I'll get back to it this weekend 😉

m90:
> Why do we have output on a normal run but not on tests? Is there a reason for that?

This is not ideal, but there is indeed a reason for that. Looking at the code:

```sh
logs=$(docker compose exec -T backup backup)

echo "$logs"
# Do some assertions on the log output
```

This means `$logs` will only be printed when the `docker compose exec -T backup backup` commands exits 0. If it does not exit 0, the script will exit immediately as `set -e` is set. Thus, the line where logs would actually get printed is never reached.

Maybe it'd be a good time now to fix this and use some `tee` (as described here: https://unix.stackexchange.com/a/724029/437580)

```
```sh
logs=$(docker compose exec -T backup backup | tee /dev/stderr)
# Do some assertions on the log output
```

---

Let me know if there are any open questions still, happy to help out.

luizguilhermesj:
_FILE support fixed!

Please let me know if there's something else that I missed or could be better =]

PR #643, opened 2025-10-04
Use forked and active shoutrrr package

### LINKED ISSUE #642: Use maintained version of shoutrrr
It seems the [original shoutrrr package](https://github.com/containrrr/shoutrrr) which is used for notifications is not seeing much activity anymore. Other consumers have apparently moved to a maintained fork https://github.com/nicholas-fedor/shoutrrr by now, which this project should do as well.

I am not sure how much API changes this brings, so someone will need to:
a. swap the dependency and fix any compiler issues that may come
b. thoroughly read the changelogs for anything that might affect users of this image

m90:
Closes #642


m90:
I'll defer merging this until the next release in order not to "break" documentation. If anyone wants/needs features that would only be available in the latest version of shoutrrr, please leave a comment and I can cut a release just for that.

PR #646, opened 2025-10-07
Ensure destination exists when using SSH storage

estebanthi:
When using SSH storage, ensure destination directory exists.


m90 (commented):
This is a nice addition, thanks. There's one thing to fix still (see comment inline), but other than that, I'm happy to ship this.

estebanthi:
Done!

PR #654, opened 2025-10-16
Add optional jittering before any backup actions

estebanthi:
Adds a new configuration option `BACKUP_JITTER` to introduce a random delay before starting any backup actions.

For context I'm backing up hundreds of containers from many servers to a single backup host with SSH on the same schedule. Handling a lot of SSH connections would be possible but would require changes I don't want either server-side or client-side. Adding a configurable startup jitter is the simplest option. I tried using hooks (`sleep $((RANDOM % $DELAY)`) but `pre-archive` runs after establishing the SSH connection, so it didn’t reduce the initial handshake load.

### DIFF
--- a/cmd/backup/run_script.go
+++ b/cmd/backup/run_script.go
@@ -57,7 +57,7 @@ func runScript(c *Config) (err error) {
 		max := s.c.BackupJitter
 		delay := time.Duration(rand.Int63n(int64(max) + 1))
 		if delay > 0 {
-			s.logger.Info("applying startup jitter", "delay", delay)
+			s.logger.Info(fmt.Sprintf("Applying startup jitter of %v", delay))
 			time.Sleep(delay)
 		}
 	}


cmd/backup/script.go
```diff
@@ -73,6 +74,15 @@ func newScript(c *Config) *script {
...
 func (s *script) init() error {
+	if s.c != nil && s.c.BackupJitter > 0 {
+		max := s.c.BackupJitter
+		delay := time.Duration(rand.Int63n(int64(max) + 1))
+		if delay > 0 {
+			s.logger.Info("applying startup jitter", "delay", delay)
+			time.Sleep(delay)
+		}
+	}
+
```
m90: The logic  in here is fine, but I find it a bit surprising to have the `init` function itself sleep in certain cases. Could you move this a level up to `runScript` where the entire flow of procedures is orchestrated?

estebanthi:
Done! Let me know if there is anything else to adjust.

PR #699, opened 2025-12-21
Add checks for tools in test suite

le-lenn:
This PR aims to make the test suite easier to use.

- Adds optional `gfind` support for macos users
- Checks if tools are present early and fails with helpful error messages if not


le-lenn:
Closed in favor of https://github.com/offen/docker-volume-backup/pull/702

PR #700, opened 2025-12-21
Use isolated docker config in test sandbox

le-lenn:
The test sandbox was mounting `${HOME}/.docker/config.json` into the manager/worker containers. That made the tests depend on my local Docker config, which was not compatible with the test suite.

This change sets `DOCKER_CONFIG=/tmp/docker-config` and mounts an empty named volume, keeping the test environment hermetic and consistent across machines.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -23,6 +23,8 @@ jobs:
         uses: docker/setup-buildx-action@v2
 
       - name: Run Tests
+        env:
+          DOCKER_CONFIG_FILE: /home/runner/.docker/config.json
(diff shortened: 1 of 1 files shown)


le-lenn:
That sounds like a great approach and would make the project easier for new developers to navigate and contribute to.

Do you want to take over or should I implement this?

m90:
> Do you want to take over or should I implement this?

Whatever you prefer. I won't get around to doing it before next year though, but I guess that's fine too.

m90:
I added another commit that applies the correct value in CI still, which seems to work. Thanks again.

PR #703, opened 2025-12-29
refactor: for 'docker compose up' use built-in flag --wait instead of 'sleep 5'

immanuwell:
## Changes
Changes in run.sh files from test/

Just use built-in flag --wait for `docker compose up`  instead of `sleep 5` - it's more reliable in case if containers won't be up after 5 seconds

@m90 fyi

### DIFF
--- a/test/age-passphrase/run.sh
+++ b/test/age-passphrase/run.sh
@@ -8,8 +8,7 @@ current_test=$(basename "$(pwd)")
 
 export LOCAL_DIR="$(mktemp -d)"
 
-docker compose up -d --quiet-pull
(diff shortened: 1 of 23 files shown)


m90:
This sounds good, but there seems to be some work to do still left:
- (some) tests now fail, `commands` is the first one, I did not check the following ones, so there might be more following. You can run them locally by invoking `./test.sh` from within the `test` directory.
- the tests seem to run slower than before now. I did not do the math, but the last CI run on this branch took 10 mins for running 7/34 test cases when current `main` takes 17 mins for running all 34 tests

PR #748, opened 2026-03-13
add tzdata package

mmichalak-swe:
## Changes
- Add `tzdata` package to Docker image

## Reason
Installing `tzdata` provides the system timezone database and allows the container timezone to be configured via the `TZ` environment variable.


mmichalak-swe:
I am not an expert in Go, I had to get AI to help me with this one. Feel free to suggest changes or modifications you think would be helpful. I think this does a good job of capturing what we need to do, which is to detect any of the bind mounts, print a deprecation message, then hand execution off to the main binary. We could add a direct link to the [page in the documentation](https://offen.github.io/docker-volume-backup/how-tos/set-container-timezone.html), to the deprecation warning.

Could this possibly be included in the existing [main.go](https://github.com/offen/docker-volume-backup/blob/main/cmd/backup/main.go) file, so that there is no need to modify the container's entrypoint? That would simplify the script, and the Dockerfile would remain the same after this functionality is added.

```Go
package main

import (
	"bufio"
	"fmt"
	"os"
	"strings"
	"syscall"
)

func main() {
	deprecatedMounts := []string{"/etc/timezone", "/etc/localtime", "/usr/share/zoneinfo"}
	foundDeprecated := false

	file, err := os.Open("/proc/self/mountinfo")
	if err == nil {
		defer file.Close()
		scanner := bufio.NewScanner(file)
		
		for scanner.Scan() {
			line := scanner.Text()
			for _, mnt := range deprecatedMounts {
				if strings.Contains(line, " "+mnt+" ") {
					fmt.Printf("WARNING: Bind-mounting '%s' is DEPRECATED.\n", mnt)
					foundDeprecated = true
				}
			}
		}
	}

	if foundDeprecated {
		fmt.Println("%%%%% DEPRECATION WARNING %%%%%")
		fmt.Println("Use the 'TZ' environment variable to set the container's time zone")
		fmt.Println("Please see the documentation for more details")
		fmt.Println("%%%%% END DEPRECATION WARNING %%%%%")
	}

	binary := "/usr/bin/backup"
	args := []string{"backup", "-foreground"}

	// syscall.Exec replaces this Go process with the backup process (PID 1)
	err = syscall.Exec(binary, args, os.Environ())
	if err != nil {
		fmt.Printf("Failed to execute %s: %v\n", binary, err)
		os.Exit(1)
	}
}
```

mmichalak-swe:
@m90 sorry for the delayed response, I have been busy, but will be able to complete this in the next few days.

m90:
No worries, please proceed as it fits your schedule and just ping me when you need help or you're done.

mmichalak-swe:
@m90 please review my latest changes when you get a chance and let me know what you think. I built the image locally and tested with the following docker-compose file:

```yaml
services:
  volume-backup:
    cap_drop:
      - ALL
    container_name: volume-backup
    environment:
      - BACKUP_CRON_EXPRESSION=0 0 5 31 2 ?
      - TZ=America/New_York
    hostname: volume-backup
    image: <local test image>
    read_only: true
    restart: unless-stopped
    security_opt:
      - no-new-privileges:true
    tmpfs:
      - /tmp
      - /var/lock
    user: "911:911"
    volumes:
      - /etc/timezone:/etc/timezone:ro
      - /etc/localtime:/etc/localtime:ro
      - /usr/share/zoneinfo:/usr/share/zoneinfo:ro
```

For reference, my current local timezone is `America/Los_Angeles` (PDT)
Here are the logs when running the above configuration:

```text
% docker logs volume-backup     
time=2026-04-15T02:49:32.568-04:00 level=WARN msg="Deprecated timezone bind mounts detected: /etc/timezone, /etc/localtime, /usr/share/zoneinfo. Support for these will be removed in a future version."
time=2026-04-15T02:49:32.568-04:00 level=WARN msg="`TZ=America/New_York` is set, but deprecated timezone bind mounts are still present. Remove the bind mounts after confirming timezone handling works as expected."
time=2026-04-15T02:49:32.568-04:00 level=INFO msg="Successfully scheduled backup from environment with expression 0 0 5 31 2 ?"
time=2026-04-15T02:49:32.568-04:00 level=WARN msg="Scheduled cron expression 0 0 5 31 2 ? will never run, is this intentional?"

% docker exec volume-backup date
Wed Apr 15 02:49:37 EDT 2026
```

When I comment out the TZ variable, the time zone in the container is still set (as expected) to my host's time zone via the bind mounts. The logs change to the following:

```text
% docker logs volume-backup     
time=2026-04-14T23:49:51.913-07:00 level=WARN msg="Deprecated timezone bind mounts detected: /etc/timezone, /etc/localtime, /usr/share/zoneinfo. Support for these will be removed in a future version."
time=2026-04-14T23:49:51.913-07:00 level=WARN msg="Set the container timezone using the `TZ` environment variable instead."
time=2026-04-14T23:49:51.913-07:00 level=WARN msg="Refer to the documentation for migration details."
time=2026-04-14T23:49:51.913-07:00 level=INFO msg="Successfully scheduled backup from environment with expression 0 0 5 31 2 ?"
time=2026-04-14T23:49:51.913-07:00 level=WARN msg="Scheduled cron expression 0 0 5 31 2 ? will never run, is this intentional?"

% docker exec volume-backup date
Tue Apr 14 23:49:55 PDT 2026
```

docs/how-tos/set-container-timezone.md
```diff
@@ -7,20 +7,49 @@ nav_order: 8
...
+>       - data:/backup/my-app-backup:ro
+>       - /etc/timezone:/etc/timezone:ro
+>       - /etc/localtime:/etc/localtime:ro
+>       - /usr/share/zoneinfo:/usr/share/zoneinfo:ro
+>
+> volumes:
+>   data:
+> ```
+
+## Recommended approach (using `TZ`)
```
m90: I would guess we can move the recommended approach up? Ideally, users will only read this far and move on without even knowing the old approach still exists?

docs/.gitignore
```diff
@@ -1,2 +1,3 @@
 _site
 .jekyll-cache
+.DS_Store
```
m90: Can we remove this? I would argue OS-specific files like this should go in a global gitignore on individual machines.

cmd/backup/command.go
```diff
@@ -150,3 +153,9 @@ func (c *command) must(err error) {
 		os.Exit(1)
 	}
 }
+
+func (c *command) logStartupWarnings(config *Config) {
```
m90: Is this method really needed? I'm a bit confused this is used in here, but then it's not used when printing the config. Maybe it'd be better to just loop over the warnings in place above, even if it's a slight repetition?

cmd/backup/config.go
```diff
@@ -322,3 +324,81 @@ func (c *Config) resolve() (reset func() error, warnings []string, err error) {
...
+		return nil, err
+	}
+
+	return mounts, nil
+}
+
+func (c *Config) timezoneDeprecationWarnings() []string {
+	mounts, err := mountedPaths("/proc/self/mountinfo")
+	if err != nil {
+		return nil
```
m90: I would think this means something went really wrong and we might not even want to continue at all? I.e. I think the method should return `[]string, error`

cmd/backup/config.go
```diff
@@ -322,3 +324,81 @@ func (c *Config) resolve() (reset func() error, warnings []string, err error) {
...
+	}
+
+	if err := scanner.Err(); err != nil {
+		return nil, err
+	}
+
+	return mounts, nil
+}
+
+func (c *Config) timezoneDeprecationWarnings() []string {
```
m90: Maybe this can have a more descriptive name? Something like `checkTimezoneConfiguration`. What do you think?

m90 (commented):
This looks good, thank you. I left a couple of comments inline, but nothing major. Let me know what you think.

docs/how-tos/set-container-timezone.md
mmichalak-swe, line 1: @m90 let me know your thoughts on the new structure of this file. I also propose that we change the tag from `v2` to `latest` as a placeholder for the docker-compose file code snippets.

m90: The docs are great now, thank you.

I would like to stick to using `v2` though as it's basically "latest" plus it gives you the guarantee that snippets flying around the internet keep working in case a v3 is ever released: If someone (or a LLM) copies portions of the current documentation, keeps them elsewhere, and then starts using them at a point when `latest` has moved on to something `v3`, things might break.

I.e. `v2` should give us the benefits of `latest` (not having to update docs over and over again, some sort of autoupdate) without the downsides.

m90 (approved):
This looks good now, thank you! I'm a bit in a hurry right now, but I'll make sure to get this merged and released next week 🎩

m90:
This is now released in [v2.48.0](https://github.com/offen/docker-volume-backup/releases/tag/v2.48.0)

m90:
For the record, Docker Hub reports the new version is 0.7MB larger than the previous, which I guess is just fine.

mmichalak-swe:
Excellent! Glad we got this done

Issue #804, opened 2026-06-17
Reduce memory usage of the SSH backend


J4kubS:
**Is your feature request related to a problem? Please describe.**
Recently some of my `docker-volume-backup` sidecars started getting killed by OOM. The issue seems to be a combination of low container memory limits (128 MiB), size of the backup archive, and buffer size used by the SSH backend. I've resolved this by increasing the memory limits relative to the archive size.

**Describe the solution you'd like**
Instead of copying the archive by sequentially reading a large (1 GiB) chunk of it into memory and then writing it to the remote file, consider using [File.ReadFrom](https://pkg.go.dev/github.com/pkg/sftp#File.ReadFrom). I've found the recommendation on [File.Write](https://pkg.go.dev/github.com/pkg/sftp#File.Write) docs while digging through the source.

At a glance it seems that all the other backends, except Dropbox, upload the archive by some form of streaming. This would make the SSH backend behave similarly.

Alternatively, using [io.Copy](https://pkg.go.dev/io#Copy) or drastically reducing the buffer size (<1 MiB) might be easier to implement.

**Describe alternatives you've considered**
Increasing the memory limits to at least 1 GiB and staggering individual backups by using `BACKUP_JITTER`, or individual schedules.

**Additional context**
I've started with the 128 MiB limits because the container idles around ~10 MiB, so it seemed to be more than enough of a margin

m90:
This sounds reasonable. Since you already figured out what needs to be done, would you mind opening a PR?

J4kubS:
Absolutely. I'll look into it tonight

J4kubS:
Addressed in #805 by using `io.Copy` for uploads

PR #805, opened 2026-06-17
Use io.Copy for SSH uploads

J4kubS:
Addresses #804

Tested with a ~300 MiB archive uploaded to an SSH backend by a container with a 64 MiB memory limit. Memory usage hovered around 35 MiB for the duration of the upload

### DIFF
--- a/internal/storage/ssh/ssh.go
+++ b/internal/storage/ssh/ssh.go
@@ -124,6 +124,12 @@ func (b *sshStorage) Copy(file string) (returnErr error) {
 		returnErr = errors.Join(returnErr, source.Close())
 	}()
 
+	sourceFileInfo, err := source.Stat()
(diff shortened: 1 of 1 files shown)


m90:
This is very nice both in terms of functionality and also as a simplification :tophat:

m90:
I'll release this once #784 is ready, which I hope is rather soon. If this not being released is blocking you lmk and I can cut a release anyways.

J4kubS:
I have temporarily raised memory limits to prevent OOMs. That's going to hold nicely until a regular release; no need for an extra one

m90:
This is now included in [v2.48.2](https://github.com/offen/docker-volume-backup/releases/tag/v2.48.2), thanks for your contribution :tophat:

PR #842, opened 2026-08-25
Add one-of matching for stop-during-backup labels

### LINKED ISSUE #621: Allow for label values that match multiple selectors
### Discussed in https://github.com/offen/docker-volume-backup/discussions/620

<div type='discussions-op-text'>

<sup>Originally posted by **huyz** August 15, 2025</sup>
If I have a list of labels like `docker-volume-backup.stop-during-backup=label1` and `docker-volume-backup.stop-during-backup=label2`, would either label cause the container to stop.  I'm not sure how valid it is to have labels with the same key</div>

---

How it could be done:

- come up with a good new label name whose value is being split on commas automatically
- keep the old behavior for the old label name, but print a deprecation warning on usage
- remove all mentions of the old label from the docs
- on the next major version, remove the old label entirely

or:
- introduce a `LABEL_MATCH_BEHAVIOR` which defaults to `MATCH` which is what is happening right now
- when this is set to `ONE_OF`, start splitting values on commas

I'm only 98% sure this can be done right now, as filtering by labels is rather restricted in Docker and adding this feature might mean we'd have to reimplement matching from scratch ourselves.

ChrisJr404:
Closes #621

Right now a container's `docker-volume-backup.stop-during-backup` label value has to match the configured `BACKUP_STOP_DURING_BACKUP_LABEL` value exactly. Since a Docker label can only hold one value per key, a single container cannot be picked up by two backup instances that use different values.

This adds the `LABEL_MATCH_BEHAVIOR` option you sketched out as the second variant in the issue. It defaults to `match`, which keeps the current exact behavior. When set to `one-of` the container's label value is split on commas, so labeling a container `stop-during-backup=service1,service2` lets an instance configured with either `service1` or `service2` stop it.

Matching already happens client side (we list all containers and filter in Go), so the Docker label filtering restrictions don't come into play here. It applies to both the `stop-during-backup` and `stop-during-backup-no-restart` labels. Added a unit test for the matcher and documented the option in the reference and the how-to.

### DIFF
--- a/cmd/backup/stop_restart.go
+++ b/cmd/backup/stop_restart.go
@@ -8,6 +8,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"strings"
 	"sync"
 	"time"
 
@@ -85,14 +86,25 @@ func isSwarm(c interface {
 	return result.Info.Swarm.LocalNodeState != "" && result.Info.Swarm.LocalNodeState != swarm.LocalNodeStateInactive && result.Info.Swarm.ControlAvailable, nil
 }
 
-func hasLabel(labels map[string]string, key, value string) bool {
+func hasLabel(labels map[string]string, key, value string, matchBehavior MatchBehavior) bool {
 	val, ok := labels[key]
-	return ok && val == value
+	if !ok {
+		return false
+	}
+	if matchBehavior == "one-of" {
+		for _, candidate := range strings.Split(val, ",") {
+			if strings.TrimSpace(candidate) == value {
+				return true
+			}
+		}
+		return false
+	}
+	return val == value
 }
 
-func checkStopLabels(labels map[string]string, stopDuringBackupLabelValue string, stopDuringBackupNoRestartLabelValue string) (bool, bool, error) {
-	hasStopDuringBackupLabel := hasLabel(labels, "docker-volume-backup.stop-during-backup", stopDuringBackupLabelValue)
-	hasStopDuringBackupNoRestartLabel := hasLabel(labels, "docker-volume-backup.stop-during-backup-no-restart", stopDuringBackupNoRestartLabelValue)
+func checkStopLabels(labels map[string]string, stopDuringBackupLabelValue string, stopDuringBackupNoRestartLabelValue string, matchBehavior MatchBehavior) (bool, bool, error) {
+	hasStopDuringBackupLabel := hasLabel(labels, "docker-volume-backup.stop-during-backup", stopDuringBackupLabelValue, matchBehavior)
+	hasStopDuringBackupNoRestartLabel := hasLabel(labels, "docker-volume-backup.stop-during-backup-no-restart", stopDuringBackupNoRestartLabelValue, matchBehavior)
 	if hasStopDuringBackupLabel && hasStopDuringBackupNoRestartLabel {
 		return hasStopDuringBackupLabel, hasStopDuringBackupNoRestartLabel, errwrap.Wrap(nil, "both docker-volume-backup.stop-during-backup and docker-volume-backup.stop-during-backup-no-restart have been set, cannot continue")
 	}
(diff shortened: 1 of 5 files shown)


docs/reference/index.md
```diff
@@ -555,6 +555,14 @@ If you need to confirm what the container actually loaded, see [Show loaded conf
 # skips restarting the container or service once the backup has finished.
 # BACKUP_STOP_DURING_BACKUP_NO_RESTART_LABEL="true"
 
+# Controls how a container's stop-during-backup label value is matched against
+# the value configured above. The default "match" requires the values to be
+# equal. Setting this to "one-of" splits the container's label value on commas,
+# so a single container labeled `stop-during-backup=service1,service2` can be
+# stopped by multiple instances of this image, each configured with a different
+# value.
+# LABEL_MATCH_BEHAVIOR="match"
```
m90: I know it's a bit nonsensical by now, but all other label related settings are prefixed `BACKUP`, i.e. I would expect this to be `BACKUP_LABEL_MATCH_BEHAVIOR`. What do you think?

m90 (commented):
Thank you very much for working on this. Two rather theoretical questions from my end:

1. Would you think we should make the separator that is used for delimiting label values configurable too?
2. Is it possible to escape a comma (or any other separator if configurable) in a `one-of` label? I.e. can I have a label value be split into `foo` and `bar,baz`?

ChrisJr404:
Renamed it to `BACKUP_LABEL_MATCH_BEHAVIOR` so it lines up with the other label settings, good call.

On the two questions:

1. I kept the separator as a fixed comma for now. Making it configurable is easy to layer on later, but I'd rather not add a second knob before there's a concrete case that needs it.
2. There's no escaping at the moment, so the value is always split on every comma. `foo,bar,baz` becomes three members, and you can't currently express a single member that itself contains a comma. If that turns out to matter we could add an escape like `\,` or the configurable separator from question 1, but my inclination is to keep it simple until someone actually hits the limit.

m90:
My reasoning here is that I don't have any control over the consumers of the image, which means that if any of these additional options require a change in behavior, I can't add them anymore. This is a fundamental difference to an application project, where your argument makes perfect sense as you'd be the only consumer and can break things more easily. Hence, adding them might still make sense here, just so there is a set of behaviors and an interface that likely won't change anytime soon.

I'd think making the separator configurable should be rather simple, I don't know about escaping though?

ChrisJr404:
Good questions.

1. Making the separator configurable is easy to add. I would keep the current comma as the default and expose an override so existing setups are unaffected.
2. Escaping is the harder part. Rather than invent an escaping scheme, I would lean on making the separator configurable so a value that contains a comma can pick a different delimiter that does not collide. Full quoting/escaping of an arbitrary separator inside a label value gets fiddly for little real-world gain, so I would leave it out unless a concrete need shows up.

On naming: agreed, `BACKUP_LABEL_MATCH_BEHAVIOR` fits the existing `BACKUP_*` label settings better. I will rename it to that for consistency.

m90:
Ok, then let's make the separator configurable, shall we? Like this, you could also work around the need for escaping I guess, so we should be all good.

ChrisJr404:
Added a configurable separator: `BACKUP_LABEL_MATCH_SEPARATOR`, defaulting to a comma. When `one-of` matching is enabled the container label value is split on that separator, so if a value itself contains a comma you can switch to a different delimiter to avoid the collision (which sidesteps the escaping question). An empty separator falls back to a comma. Threaded it through `hasLabel`/`checkStopLabels`, added test cases for a custom separator and the empty fallback, and documented it in the reference and the how-to. The env var rename to `BACKUP_LABEL_MATCH_BEHAVIOR` from earlier is in as well.

cmd/backup/stop_restart.go
```diff
@@ -85,14 +86,28 @@ func isSwarm(c interface {
...
 
-func hasLabel(labels map[string]string, key, value string) bool {
+func hasLabel(labels map[string]string, key, value string, matchBehavior MatchBehavior, separator string) bool {
 	val, ok := labels[key]
-	return ok && val == value
+	if !ok {
+		return false
+	}
+	if matchBehavior == "one-of" {
+		if separator == "" {
```
m90, line 95: I do see how this is trying to introduce convenience, but I'm not sure it is the expected behavior. There might be users that set an empty string on purpose ("exploding" the string). Even if this use case sounds nonsensical, I would argue it's better to support it if it's easily possible. It's also not mentioned in the docs, so it might get confusing.

Another option would be bailing with a meaningful error message here, but I guess that's not needed either.

m90 (commented):
Thanks for the extra effort you put into this. I'd like to remove the fallback for an empty separator value again still, as it's slightly too much magic, but other than that, this looks fantastic and will be a great addition to ship.

m90:
I've got some time on my hands today, so I'll merge this and remove the fallback myself. I hope you don't mind. Thanks for adding this feature :tophat:

m90:
This is now released in [v2.49.0](https://github.com/offen/docker-volume-backup/releases/tag/v2.49.0), thank you :tophat:

PR #848, opened 2026-09-06
Support tencent cloud COS

sainnhe:
Some S3 providers (e.g. Tencent Cloud COS) requires custom bucket lookup method. This PR adds a new env var named `AWS_S3_BUCKET_LOOKUP` to support this feature.

I've confirmed that it works with the following setup:

- Image: `ghcr.io/sainnhe/docker-volume-backup:v2`
- Envs:

```
AWS_ENDPOINT="cos.ap-guangzhou.myqcloud.com"
AWS_ENDPOINT_PROTO="https"
AWS_S3_BUCKET_LOOKUP="dns"
AWS_S3_BUCKET_NAME="xxx"
AWS_ACCESS_KEY_ID="AKxxx"
AWS_SECRET_ACCESS_KEY="xxx"
```

### DIFF
--- a/internal/storage/s3/s3.go
+++ b/internal/storage/s3/s3.go
@@ -38,6 +38,7 @@ type Config struct {
 	BucketName       string
 	StorageClass     string
 	PartSize         int64
+	BucketLookup     string
 	CACert           *x509.Certificate
 }
 
@@ -59,6 +60,16 @@ func NewStorageBackend(opts Config, logFunc storage.Log) (storage.Backend, error
 	options := minio.Options{
 		Creds:  creds,
 		Secure: opts.EndpointProto == "https",
+	}
+	switch opts.BucketLookup {
+	case "dns", "virtual":
+		options.BucketLookup = minio.BucketLookupDNS
+	case "path":
+		options.BucketLookup = minio.BucketLookupPath
+	case "", "auto":
+		options.BucketLookup = minio.BucketLookupAuto
+	default:
+		return nil, errwrap.Wrap(nil, fmt.Sprintf("unknown AWS_S3_BUCKET_LOOKUP value: %s", opts.BucketLookup))
 	}
 
 	transport, err := minio.DefaultTransport(true)
--- a/internal/storage/s3/s3.go
+++ b/internal/storage/s3/s3.go
@@ -66,7 +66,7 @@ func NewStorageBackend(opts Config, logFunc storage.Log) (storage.Backend, error
 		options.BucketLookup = minio.BucketLookupDNS
 	case "path":
 		options.BucketLookup = minio.BucketLookupPath
-	case "", "auto":
+	case "auto":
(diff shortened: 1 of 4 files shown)


internal/storage/s3/s3.go
```diff
@@ -60,6 +61,16 @@ func NewStorageBackend(opts Config, logFunc storage.Log) (storage.Backend, error
...
 	}
+	switch opts.BucketLookup {
+	case "dns", "virtual":
+		options.BucketLookup = minio.BucketLookupDNS
+	case "path":
+		options.BucketLookup = minio.BucketLookupPath
+	case "", "auto":
+		options.BucketLookup = minio.BucketLookupAuto
+	default:
+		return nil, errwrap.Wrap(nil, fmt.Sprintf("unknown AWS_S3_BUCKET_LOOKUP value: %s", opts.BucketLookup))
```
m90, line 72: In the configuration mechanism, there is a `resolve` method: https://github.com/offen/docker-volume-backup/blob/bd50ba96c439e6003b982cf6b371278eff946669/cmd/backup/config.go#L239 which makes sure the config is "workable" for the rest of the program. Ideally, checking whether the given lookup value is valid should already happen there.

Alternatively, you could also use a custom type to emulate an enum as currently happening over here: https://github.com/offen/docker-volume-backup/pull/842/changes#diff-0d74e7cb064512fb300977d065a22b258d5b6d9f40d17ca211318c135b331364R125-R146

internal/storage/s3/s3.go
```diff
@@ -60,6 +61,16 @@ func NewStorageBackend(opts Config, logFunc storage.Log) (storage.Backend, error
 		Creds:  creds,
 		Secure: opts.EndpointProto == "https",
 	}
+	switch opts.BucketLookup {
+	case "dns", "virtual":
+		options.BucketLookup = minio.BucketLookupDNS
+	case "path":
+		options.BucketLookup = minio.BucketLookupPath
+	case "", "auto":
```
m90: I would think handling the empty string as an invalid value would feel more obvious. Since there is a default value of `auto`, users will need to do something to set the empty string, so they probably want something different than the default behavior. Raising an error will be the easiest path for them to find out what's wrong.

m90 (commented):
Thanks for your contribution. I left some minor comments inline, let me know what you think.

internal/storage/s3/s3.go
sainnhe: Makes sense. I've fixed it in the latest commit.

I'm using the first approach since it organizes all checks in one place.

internal/storage/s3/s3.go
sainnhe: Makes sense. I've fixed it in the latest commit.

m90:
This is now released in [v2.49.0](https://github.com/offen/docker-volume-backup/releases/tag/v2.49.0), thank you :tophat:

